PR #4, opened 2022-02-02
Solution

alix-mak:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str):
+    split_command = command.split()
+    if split_command[1] != split_command[2]:
+        with open(split_command[1], 'r') as f1:
```
IvanRamyk, line 4: Let's use more descriptive file names

IvanRamyk, line 4:
```suggestion
        with open(split_command[1], 'r') as source:
```

app/main.py
```diff
@@ -1 +1,9 @@
...
+            lines = f1.readlines()
+            with open(split_command[2], "w") as f2:
```
IvanRamyk, line 6:
```suggestion
            with open(split_command[2], "w") as destination:
```

IvanRamyk (changes requested):

PR #8, opened 2022-02-02
'Solution'

liuda-chuzdiuk:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str):
+    sep_command = command.split()
+    file = sep_command[1]
+    new_file = sep_command[2]
+
+    if file != new_file:
+        with open(file, "r") as f:
```
IvanRamyk, line 7:
```suggestion
        with open(file, "r") as source:
```

IvanRamyk (changes requested):

app/main.py
alix-mak, line 7: Let's use more descriptive file names. For example, source instead of f.

alix-mak (changes requested):

IvanRamyk (approved):
Well done!!

PR #12, opened 2022-02-02
Solution

vldtsw:


alexandra-halytska (approved):

app/main.py
```diff
@@ -1 +1,10 @@
...
+import shutil
+
+
+def copy_file(command: str):
+    new_command = command.split()
+    file = new_command[1]
+    new_file = new_command[2]
+
+    if file != new_file:
+        shutil.copy(file, new_file)
```
IvanRamyk, line 10: In practice, you can use this module
But the task was to copy a file using file.read and file.write methods,

IvanRamyk (changes requested):

vldtsw:
Had some problems with git in PyCharm :)

PR #13, opened 2022-02-02
'solution'

alexandra-halytska:


app/main.py
```diff
@@ -1 +1,4 @@
-# write your code here
+def copy_file(command):
```
IvanRamyk, line 1: This code looks like magic to an outsider)
Could you create intermediate variables to make it more readable)

app/main.py
```diff
@@ -1 +1,4 @@
...
+    with open(command.split(' ')[1], "r") as f1:
```
IvanRamyk, line 2: Let's use more descriptive names for the file variables

```suggestion
    with open(command.split(' ')[1], "r") as source:
```

app/main.py
```diff
@@ -1 +1,4 @@
...
+        with open(command.split(' ')[2], "w") as f2:
```
IvanRamyk, line 3:
```suggestion
        with open(command.split(' ')[2], "w") as destination:
```

IvanRamyk (changes requested):

IvanRamyk (changes requested):
Consider the case when the source name is equal to the destination name
```
copy text.txt text.txt
```

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command):
+    with open(command.split(' ')[1], "r") as source:
+        with open(command.split(' ')[2], "w") as destination:
+            if command.split(' ')[1] != command.split(' ')[2]:
```
IvanRamyk, line 4: It's better to check this condition before you actually open files

IvanRamyk (changes requested):

IvanRamyk (approved):
Good job!

PR #16, opened 2022-02-02
'Solution'

8Men5Camels:


vldtsw (approved):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command):
+    data = command.split()
+    if data[0] == 'cp' and data[1] != data[2]:
+        with open(data[1], 'r') as source:
+            with open(data[2], 'w') as result:
+                result.write(str(source.read()))
```
IvanRamyk, line 6:
```suggestion
                result.write(source.read())
```
`source.read` returns a string type. You don't need to convert the result to a string

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

IvanRamyk (approved):
Nice job!

PR #18, opened 2022-02-02
'Solution'

Oleksandr563487953:


denyslipin (approved):
A few suggestions:
1. you can unpack "command" in 1 line: file, file_copy = (through a slice)
or even: [main_command, file, file_copy = ...]  and also check if main_command is right
2. method split by default use whitespace as a separator: split() == slit(" ")
3. you don't need to create "text" variable, just put origin.read() in copy.write()

IvanRamyk (approved):
Good job!

PR #25, opened 2022-05-12
Solution

KatiukhaO:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import re
+
+
+def copy_file(command: str):
+    file_names = re.findall(r"(\w+[-_.,]?\w+.txt)", command)
```
MasterpieceElbow, line 5: It is a good solution. But it is not very good to use regex when you can do without it. Refactor.

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    if file_names[0] != file_names[1]:
+        with open(file_names[0], "r") as f:
+            with open(file_names[1], "w") as f2:
+                f2.write(f.read())
+                print("File copy is created successes")
+
+
+copy_file("cp test.txt test-copy.txt")
```
MasterpieceElbow, line 14: Remove unnecessary print and function call.

MasterpieceElbow (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
...
+def copy_file(command: str):
+    file_names = command.split(" ")
+
+    if file_names[1] != file_names[2]:
+        with open(file_names[1], "r") as f:
+            with open(file_names[2], "w") as f2:
+                f2.write(f.read())
+
+
+copy_file("cp test.txt test-copy.txt")
```
MasterpieceElbow, line 10: Remove function call.

MasterpieceElbow (changes requested):

TPolina (approved):

PR #32, opened 2022-05-29
Solution

Magus6784:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str):
+    command = command.split()
+    if command[0] == "sp" and command[1] != command[2]:
```
TPolina, line 3: The correct command for copying is `cp` not `sp`

TPolina (changes requested):

TPolina (approved):
Good!

PR #33, opened 2022-05-30
'Solution'

Korenely:


katevasilenko (approved):
I have no english words for that, so: А шо, так можна було?

TPolina (approved):
You were supposed to practice opening, reading, writing files, but ok :)))

PR #41, opened 2022-05-31
Katya's Solution

katevasilenko:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str):
+    check_list = command.split(" ")
```
lenasharabura, line 2: split automatically splits on spaces, no need to use " "

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if check_list[1] != check_list[2]:
+        with open(f"{check_list[1]}", "r") as file_in:
```
lenasharabura, line 4: no need to use f-formatting, check_list[1] is a string anyway
use just check_list[1]

app/main.py
```diff
@@ -1 +1,9 @@
...
+            lines = file_in.readlines()
+
+        with open(f"{check_list[2]}", "w") as file_out:
```
lenasharabura, line 7: same

lenasharabura (approved):
good job. just some notes

app/main.py
katevasilenko: Thank you for useful advice

TPolina (approved):

PR #44, opened 2022-05-31
Solution

TsybuliaRoman:


katevasilenko (approved):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str):
+    command_list = list(command.split())
+    if command_list[0] == "cp" and command_list[1] != command_list[2]:
+        with open(f"{command_list[1]}", "r") as old_file:
+            with open(f"{command_list[2]}", "w") as new_file:
```
TPolina, line 5: there is no need to use f-string

PR #49, opened 2022-05-31
Solution

Chardona:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str):
+    cmdList = command.split()
```
rotsen18, line 2: use snake_case for naming variables
[PEP](https://peps.python.org/pep-0008/#function-and-variable-names)

rotsen18 (changes requested):

app/main.py
Chardona: habit to call in camel Case ;)

TPolina (approved):

PR #52, opened 2022-05-31
Solution

Evgeniy994:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str):
+    names = command.split('')
```
katevasilenko, line 2: your separator in split method is empty. Just delete any quotes, split() use space by default

katevasilenko (changes requested):

MasterpieceElbow (approved):

PR #53, opened 2022-05-31
'Solution'

Kiawami:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+# If you're reviewing this code - thank you <3
+
+def copy_file(command: str):
+    command_list = list(command.split(" "))
```
katevasilenko, line 4: if you split, it already the list, and also split use space by defauld, so yoou don't need " "

katevasilenko (changes requested):

PR #57, opened 2022-06-01
Solution

Igor-Cegelnyk:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command):
+    command_list = command.split(" ")
+    with open(f"{command_list[1]}", "r") as file_in, \
```
Abnormaltype, line 3: you must check the command, and name of the file before the context manager

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command):
+    command_list = command.split(" ")
+    if command_list[1] != command_list[2]:
```
Abnormaltype, line 3: also, you should check the command `cp`

Abnormaltype (changes requested):

PR #60, opened 2022-06-03
Solution

darten360:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str):
+    lst = command.split(" ")
+    if lst[0] == "cp" and lst[1] != lst[2]:
+        with open(f"{lst[1]}", "r") as file_in, open(f"{lst[2]}", "r") as file_out:
```
rotsen18, line 4: Does this code realy working? Because I see you opened file_out for reading also.

rotsen18 (approved):
Now it seems should work. And in this case you might write split() without parameter

darten360:
> Now it seems should work. And in this case you might write split() without parameter. But line is too long.

But now it doen't work. Tests were failed

PR #70, opened 2022-06-25
Solution

CEHR2005:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str):
+    file1 = command.split()[1]
+    file2 = command.split()[2]
+    if file1 == file2:
```
Abnormaltype, line 4: unnecessary condition.

make a check, if they are not equal, then make a record

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str):
+    file1 = command.split()[1]
+    file2 = command.split()[2]
+    if file1 != file2:
+        with open(file1, "r") as file_in, open(file2, "w") as file_out:
+            if file_in.read() == file_out.read():
```
Abnormaltype, line 6: why not `if not file_in.read() == file_out.read():` ? and `file_out.write(file_in.read())`?

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str):
+    file1 = command.split()[1]
+    file2 = command.split()[2]
+    if file1 != file2:
```
Abnormaltype, line 4: also you should check input command

Abnormaltype (changes requested):

Abnormaltype (approved):
well done!

PR #73, opened 2022-07-15
solution

tarkaiev:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command):
+    old_file = command.split(" ")[1]
+    new_file = command.split(" ")[2]
+
+    if old_file == new_file:
```
Abnormaltype, line 5: also, you should check the input command

Abnormaltype (changes requested):

app/main.py
tarkaiev: updated

PR #80, opened 2022-07-22
Solution

andriy-demeshko:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command):
+    file_names = command.split()
+    print(file_names)
```
Abnormaltype, line 3: do you need this?

Abnormaltype (changes requested):
It must do nothing in case the user is trying to copy file to file with the same name.

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command):
+    file_names = command.split()
+    if file_names[1] == file_names[2]:
```
Abnormaltype, line 3: also, you should check the input command `cp`

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #81, opened 2022-07-24
Solution

Vlad-Peresta:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str):
+    command_list = command.split(" ")
+    copy_from = command_list[1]
+    copy_to = command_list[2]
+
+    if copy_from != copy_to:
```
Abnormaltype, line 6: also, you should check the input command `cp`

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

app/main.py
Vlad-Peresta: ok, thank you)!

PR #89, opened 2022-08-11
Solution

VladyslavKlevanskyi:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str):
+    file1 = command.split(" ")[1]
+    file2 = command.split(" ")[2]
+    if file1 != file2:
```
Abnormaltype, line 4: also, you should check the input command

Abnormaltype (changes requested):

app/main.py
VladyslavKlevanskyi: What exactly needs to be checked in the input command?

Abnormaltype: image

Abnormaltype (approved):
Well done!

PR #92, opened 2022-08-18
'Solution'

Viktor-Beniukh:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str):
+
+    command = command.split()
+
```
DmytroSvirsa, line 4: also, you should check the input command

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #96, opened 2022-08-21
Solution

volodymyr-vereshchak:


app/main.py
```diff
@@ -1 +1,38 @@
-# write your code here
+from app.errors import CommandError, CopyFileExistsError
+
+
+def file_copy(command, file1, file2) -> bool:
+    if command and command == 'cp':
```
Abnormaltype, line 5:
```suggestion
    if command == 'cp':
```

Abnormaltype (changes requested):
also, use `""` instead of `''`

Abnormaltype (approved):
Well done!

PR #97, opened 2022-08-23
Complete task

Glasis9:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str):
+    temp = command.split()
+    if temp[-1] == temp[-2]:
+        pass
+    else:
+        with open(f"{temp[-2]}", 'r') as file_in, \
+                open(f"{temp[-1]}", 'w') as file_out:
+            file_out.write(f"{file_in.read()}")
```
Abnormaltype, line 8:
```suggestion
    if temp[-1] != temp[-2]:
        with open(f"{temp[-2]}", "r") as file_in, \
                open(f"{temp[-1]}", "w") as file_out:
            file_out.write(f"{file_in.read()}")
```

Abnormaltype (changes requested):
also, you should check the input command

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str):
+    temp = command.split()
+    if temp[-1] != temp[-2]:
```
Abnormaltype, line 3: also, you should check the input command

Abnormaltype (changes requested):
use "" instead of ''

app/main.py
Abnormaltype: what about this?)

Abnormaltype (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str):
+    str_list_command = command.split()
+    if str_list_command[-1] != str_list_command[-2]:
```
Abnormaltype, line 3: you should check the input command

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #104, opened 2022-09-08
Solution

lksndr307:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str):
+    file = command.split(" ")
+    if file[1] != file[2]:
```
DmytroSvirsa, line 3: you should check the input command

DmytroSvirsa (changes requested):

app/main.py
lksndr307: I don't understand, but I check it, split the command, and fit.

Abnormaltype, line 3: you should add a condition for the "cp" input command

Abnormaltype (changes requested):

app/main.py
lksndr307: Sorry, but I don't get it. I have the condition.HELP)

Abnormaltype: image
check if the input command is `cp`

Abnormaltype (changes requested):

PR #107, opened 2022-09-14
Solution

vlad-milkov:


Abnormaltype (changes requested):
you should check the input command

PR #113, opened 2022-09-20
Solution

wQuelS:


AriohBelskij (approved):
Awesome!

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+
+    if command[1] == command[2]:
+        return
+
+    with open(command[1], "r") as file_in, open(command[2], "w") as file_out:
```
Nattalli, line 7: should not you to check `if command[0] == "cp"`?

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
```
Nattalli, line 2: maybe `commands` anme would be better? as it is a list, not one value

Nattalli (changes requested):

Nattalli (approved):
Awesome! 😎

PR #117, opened 2022-09-22
Solution

SorokaIgor:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str):
+    new_file = command.split(" ")
+    new_file.remove(new_file[0])
```
IvanRamyk, line 3: Why remove the command?(

app/main.py
```diff
@@ -1 +1,10 @@
...
+    print(new_file)
```
IvanRamyk, line 4: remove extra print from solution

IvanRamyk (changes requested):

app/main.py
SorokaIgor: image
i deleted "cp" from string

IvanRamyk: You don't need to delete this part of the command
Because of this your code doesn't work :( 
Screenshot 2022-09-22 at 16 04 04

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #120, opened 2022-09-23
Ready

maxitoska:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str):
+    names = command.split(" ")
```
DmytroSvirsa, line 2: check the command word

DmytroSvirsa (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str):
+    if command:
```
Abnormaltype, line 2: if `command = "del"` it's should work?

Abnormaltype (changes requested):

app/main.py
maxitoska: no

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str):
+    if command and command != "del":
```
Abnormaltype, line 2:
```suggestion
    if command == "cp":
```

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #121, opened 2022-09-25
'Solution'

AVyha:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
```
mltdll, line 2: Here I'd also name arguments appropriately, even though it's not as important as in the move-file task.

app/file.txt
```diff
@@ -0,0 +1,4 @@
+123
+23
+33
+12
```
mltdll, line 4: You didn't have to add your test file to the pull request. :)
I'd actually even delete it.

mltdll (changes requested):

mltdll (approved):

Nattalli (approved):
Awesome! 😎

PR #122, opened 2022-09-25
Solution

luxuriant777:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str):
```
Nattalli, line 1: add type annotation

Nattalli (changes requested):

app/main.py
luxuriant777: added

TPolina (approved):
Well done!

PR #130, opened 2022-09-28
Solution for the task

olia-trofymchuk:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(string: str) -> None:
+    command, from_file, to_file = string.split(" ")
+
+    if from_file != to_file and command == "cp":
+        with open(from_file, "r") as file_in, open(to_file, "w") as file_out:
+            content = file_in.read()
```
mltdll, line 6: I'd write it directly to the output file, without introducing a variable.

mltdll (changes requested):
Good solution, but I'd refractor it a bit. Just request re-review if you disagree. :)

app/main.py
olia-trofymchuk: i think it's a good remark)
thanks a lot! i'll implement it😇

mltdll (approved):

Nattalli (approved):
Good job!

PR #135, opened 2022-09-28
Complete!

ketstap162:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[0] == 'cp' and command[1] != command[2]:
+        with open(command[2], 'r') as copied_file:
```
mltdll, line 4: You can open both files at the same time, as written in the README:
```python
    ...
```
And then pass contents of the first file directly into `write()` method, without introducing an extra variable.

mltdll (changes requested):

app/main.py
ketstap162: Благодарю

mltdll (approved):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[0] == 'cp' and command[1] != command[2]:
```
Nattalli, line 3: We recommend using double quotes everywhere
Single quotes should be used if double quotes are used inside the f-string
so, please. use here double quotes

Nattalli (changes requested):

Nattalli (approved):
Good job! 😸

PR #137, opened 2022-09-29
'Solution'

rakamakaphone:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command):
+    a = command.split(" ")
```
DmytroSvirsa, line 2: a is a bad variable name

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if a[1] != a[2]:
```
DmytroSvirsa, line 3: check what command was provided

DmytroSvirsa (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command):
+    string = command.strip().split(" ")
+    file1 = string[1]
+    file2 = string[2]
+    if file1 != file2:
```
DmytroSvirsa, line 5: check is that `cp` command was provided

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #138, opened 2022-09-29
Solution

Aleksandro777:


app/test
```diff
@@ -0,0 +1,2 @@
+sdfsdf
```
mltdll, line 1: I think you should delete this file from git. :)

mltdll (changes requested):

mltdll (approved):

Nattalli (changes requested):
You have to add a functionality that the first word in the command is `cp`:
``` Function take only one argument command, that is string with com```

Nattalli (approved):
Well done! 👍🏻

PR #139, opened 2022-09-29
Solution Vadim Sabadyr

VadimSabadyr:


mltdll (changes requested):
You seem to have forgotten to check for this condition:
- It must do nothing in case the user is trying to copy file to file with the same.

app/main.py
```diff
@@ -1,4 +1,6 @@
 def copy_file(command: str):
+    if command[1] == command[2]:
```
mltdll, line 2: You've introduced it before you even split the command in parts. Move it down a little.

mltdll (changes requested):

Nattalli (approved):
Well done! 👍🏻

PR #140, opened 2022-09-29
Solution

pazyn404:


mltdll (changes requested):
Solution must meet this requirement from README: 
- It must do nothing in case the user is trying to copy file to file with the same name.

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command_line: str) -> None:
+    _, source_name, destination_name = command_line.split(" ")
+
+    if source_name != destination_name:
```
mltdll, line 4: It is usually considered best practice not to put function's main logic into an if condition. I'd suggest to rewrite the check so that the function exits `if source_name == destination_name` and proceeds to execute otherwise.

mltdll (changes requested):

mltdll (approved):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command_line: str) -> None:
+    _, source_name, destination_name = command_line.split(" ")
+
+    if source_name == destination_name:
+        return
+
+    with (open(source_name, "r") as source,
```
Nattalli, line 7: as it told in README:
`Function take only one argument command, that is string with command cp`
You should to check this condition

Nattalli (changes requested):

Y-Havryliv (approved):
GJ!

PR #141, opened 2022-09-30
'Solution'

Yurchikkk:


app/main.py
```diff
@@ -1 +1,7 @@
 # write your code here
+def copy_file(command):
+    if not command.split()[1] == command.split()[2]:
```
DmytroSvirsa, line 3: check what command is provided

DmytroSvirsa (changes requested):

app/main.py
Yurchikkk: comment is not clear, check where? tests file is empty in this task

DmytroSvirsa (approved):

PR #144, opened 2022-10-01
Solution

HStakhiv:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    old_file, new_file = command.split(" ")[1:]
+
+    if old_file != new_file:
```
mltdll, line 4: Putting the function's main logic inside an `if` condition is usually considered a bad practice.

Also, the function shouldn't copy file if the command isn't `cp`.

mltdll (changes requested):
You sho

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(line: str) -> str:
+    command, old_file, new_file = line.split(" ")[1:]
+
+    if command != "cp" or old_file == new_file:
+        return "Incorrect command or your file names are the same!"
```
mltdll, line 5: I'd say that the function shouldn't return this string. You should either use empty `return` statement or raise an error with this message.

mltdll (changes requested):

mltdll (approved):

app/main.py
HStakhiv: Thank you!

Y-Havryliv (approved):
Nice!

PR #145, opened 2022-10-01
solve the task

Cerne13:


mltdll (approved):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
```
TPolina, line 1: Did you test your task? Does it work?

TPolina (changes requested):

app/main.py
Cerne13: Terribly sorry. Somehow deleted parentheses after .split (was toying w/code a bit after testing). Fixed the issue.

TPolina (approved):
Well done!

PR #147, opened 2022-10-01
'Solution'

Menselface:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> str | None:
+    result = command.split(" ")
+    file = result[1]
+    copy_of_file = result[2]
+    if result[0] != "cp":
+        return f"{result[0]} is unidentified"
```
mltdll, line 6: I'd say that this function shouldn't return anything in any case, since it's not how you would use it. Instead, if you want to send message to the programmer, you should raise an error. 
I'd also suggest to change the message to something like `Unknown command: {result[0]}`.

mltdll (changes requested):

mltdll (approved):

Y-Havryliv (approved):
GJ!

PR #148, opened 2022-10-02
Solution

graandda:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str):
+    cmd_list = command.split()
+    cmd = cmd_list[0]
+    old_file = cmd_list[1]
+    new_file = cmd_list[2]
+
+    if old_file != new_file and cmd == "cp":
+        with open(old_file, "r") as file, open(new_file, "w") as new_file:
```
mltdll, line 8: Since this is the function's main logic, you probably shouldn't put it into an `if` block. Instead, I'd suggest to reverse the condition and put `return` statement into it:
```python
if ...:
    return

with ...

**Review by mltdll — CHANGES_REQUESTED** [review]

File: app/main.py (hunk 1 above)
Comment (graandda) [reply app/main.py:10]: thank you)

File: app/main.py [hunk 2]
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str):
+    cmd_list = command.split()
+    cmd = cmd_list[0]
+    old_file = cmd_list[1]
+    new_file = cmd_list[2]
+
+    if old_file == new_file and cmd != "cp":
```
mltdll, line 7: You'd also want to use `or` operator here, not `and`.

mltdll (changes requested):

app/main.py
graandda: thx)

mltdll (approved):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str):
+    cmd_list = command.split()
+    cmd = cmd_list[0]
+    old_file = cmd_list[1]
+    new_file = cmd_list[2]
+
+    if old_file == new_file or cmd != "cp":
+        return print("Something wrong!")
+    else:
```
Y-Havryliv, line 9: you don't need `else` here

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str):
+    cmd_list = command.split()
```
Y-Havryliv, line 2: `command` would be better

app/main.py
```diff
@@ -1 +1,11 @@
...
+    cmd = cmd_list[0]
+    old_file = cmd_list[1]
+    new_file = cmd_list[2]
+
+    if old_file == new_file or cmd != "cp":
+        return print("Something wrong!")
+    else:
+        with open(old_file, "r") as file, open(new_file, "w") as new_file:
```
Y-Havryliv, line 10: `file` and `new_file` aren't really meaningful, better name it like `source_file` and `destination_file`, or smth like that

Y-Havryliv (changes requested):

Y-Havryliv (approved):
Nice!

PR #151, opened 2022-10-02
Solution

linaSam:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str):
+    lst_separated = command.split(" ")
+    if lst_separated[0] == "cp" and lst_separated[1] != lst_separated[2]:
```
HStakhiv, line 3: It is bad practice to write the main logic inside an `if` condition

HStakhiv (changes requested):
Avoid `\`

HStakhiv (approved):

Y-Havryliv (approved):
GJ!

PR #155, opened 2022-10-02
Solution

VitaliiPiaskivskyi:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str):
+    old_file = command.split()[1]
```
pazyn404, line 2: it's not old_file rather source file name, rename to source_filename

app/main.py
```diff
@@ -1 +1,7 @@
...
+    new_file = command.split()[2]
```
pazyn404, line 3: also here, it's destination file name

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if old_file != new_file:
```
pazyn404, line 4: your should also check if command.split(" ")[0] == "cp". Refactor this block to leave from function if source_filename is equal to destination_filename or command.split(" ")[0] != "cp"

pazyn404 (changes requested):

app/main.py
VitaliiPiaskivskyi: Technically, file 1 is older than file 2. When file 1 already exists, file 2 does not exist. I like variable names like this

app/main.py
VitaliiPiaskivskyi: Please read above ↑

app/main.py
VitaliiPiaskivskyi: The tests have passed, so everything is OK! Many checks can be written: Is there a "command" in the string? does the string consist of 3 parameters separated by a space? Or maybe parameter 3 contains incorrect characters in the name? and others. The task indicated writing a working algorithm, and not implementing all possible exceptions

pazyn404: Ok, you are welcome

app/main.py
Y-Havryliv: I agree with Yurii, `old_file` and `new_file` aren't really meaningful

app/main.py
Y-Havryliv: You definitely need to check if the command is correct

Y-Havryliv (changes requested):

PR #156, opened 2022-10-03
Solution

VadymShkarbul:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd = command.split()
+    if cmd[0] == "cp" and cmd[1] != cmd[2]:
+        with open(cmd[1], "r") as source, open(cmd[2], "w") as copy:
```
skrypka-dmytro, line 4: with open(cmd[1], "r") as source, open(cmd[2], "w") as copy: 

it's bad practice

use: with open(cmd[1], "r+") as source:

skrypka-dmytro (changes requested):

app/main.py
VadymShkarbul: Thank you for review. I don't agree. In this case all we need to do with "source" - read. No need to write something there.

Y-Havryliv (approved):

PR #158, opened 2022-10-03
Solution

skrypka-dmytro:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file, file_copy = command.split(",")
```
ketstap162, line 2: I call function: `copy_file("cp old_file.txt new_file.txt")` and what i can get?

ketstap162 (changes requested):

ketstap162 (approved):
Nice!

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file, file_copy = command.split(" ")
```
Y-Havryliv, line 2: it would be better to change `cp` to `command` for example

app/main.py
```diff
@@ -1 +1,6 @@
...
+
+    if file == file_copy or cp != "cp":
+        with open(file, "r") as line, open(file, "w") as copy:
```
Y-Havryliv, line 5: choose better names for files

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file, file_copy = command.split(" ")
+
+    if file == file_copy or cp != "cp":
```
Y-Havryliv, line 4: `It must do nothing in case the user is trying to copy file to file with the same name.`

Y-Havryliv (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file, file_copy = command.split(" ")
+
+    if file != file_copy or cp == "cp":
```
Y-Havryliv, line 4: `if file == file_copy or cp != "cp":`

Y-Havryliv (changes requested):

PR #160, opened 2022-10-03
Solution

YevheniiDavydenko:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str):
+    sep_cm = command.split(" ")
+    if sep_cm[1] == sep_cm[2]:
```
HStakhiv, line 3: The function should copy a file if the command is `cp`!

HStakhiv (changes requested):
Delete `file.txt`

app/main.py
```diff
@@ -1 +1,8 @@
...
+        return
+    if sep_cm[0] == "cp":
```
HStakhiv, line 5: Place this condition in line #3

HStakhiv (changes requested):
Please delete `file.txt`

HStakhiv (approved):
Not bad!

TPolina (approved):
Well done!

PR #162, opened 2022-10-03
Solution

ShevchukVitalii:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[0] == "cp" and command[1] != command[2]:
+        with open(command[1], "r") as source_file,\
```
mltdll, line 4: I'd suggest to use parenteses instead of backslash:
```python
with (open(command[1], "r") as source_file,
      open(command[2], "w") as result_file):

mltdll (changes requested):

mltdll (approved):

Y-Havryliv (approved):
GJ!

PR #163, opened 2022-10-03
'Solution'

Ilyakson:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, old_file, new_file = command.split()
```
HStakhiv, line 2: It is better to use a different name for a variable `command` that contains only part of an entire `command`

HStakhiv (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, old_file, new_file = command.split()
+
+    if old_file == new_file or command != "cp":
```
HStakhiv, line 4: Maybe ʼcmd`, not `command`?

HStakhiv (changes requested):

HStakhiv (approved):
Good!

TPolina (approved):
Well done!

PR #164, opened 2022-10-03
Develop

serviza:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[1] == command_list[2]:
```
mltdll, line 3: You should also check if the first word of the command is `cp`.

app/main.py
```diff
@@ -1 +1,9 @@
...
+        return
+    else:
+        with open(command_list[1], "r") as file_in,\
```
mltdll, line 6: It's better to use `()` than `\` to break lilnes.

app/main.py
```diff
@@ -1 +1,9 @@
...
+                open(command_list[2], "w") as file_out:
+            lines = file_in.readlines()
```
mltdll, line 8: You may also write the output of `readlines()` directly into `writelines()` without introducing a variable, if you want.

mltdll (changes requested):

app/main.py
serviza: Thanks. Done.

app/main.py
serviza: Something like this?

mltdll (approved):

app/main.py
serviza: I can't use () in a right way in this case. It will be in another way)

mltdll: You sure can! All you have to do is to enclose everything after `with` in `()`:
```python
with (open...,
      open...)

Y-Havryliv (approved):
Well done!

PR #167, opened 2022-10-03
'Solution'

Nikyxa:


pazyn404 (approved):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str):
+    list_commands = command.split()
```
TPolina, line 2: `list_commands` is a bit strange name in my opinion

app/main.py
```diff
@@ -1 +1,12 @@
...
+    command = list_commands[0]
+    first_file = list_commands[1]
+    new_file = list_commands[2]
+
+    if first_file == new_file and command != "cp":
```
TPolina, line 7: so will you do nothing only when both conditions are `true`?

app/first_file.txt
```diff
@@ -0,0 +1 @@
+check
```
TPolina, line 1: delete unnecessary files from GitHub

TPolina (changes requested):

TPolina (approved):
Well done!

PR #171, opened 2022-10-03
copy file first attempt

VitaliyBychevoy:


app/main.py
```diff
@@ -1 +1,14 @@
 # write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
+
+    if command_list[1] == command_list[2]:
```
HStakhiv, line 5: The function should copy a file if the command is `cp`!

app/main.py
```diff
@@ -1 +1,14 @@
...
+        return
+    else:
```
HStakhiv, line 7: It is bad practice to write the main logic inside an `if` condition

HStakhiv (changes requested):

app/main.py
HStakhiv: There is no need for `else`

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
+
+    if command_list[0] != "cp" and command_list[1] != command_list[2]:
```
HStakhiv, line 4: It won't work

HStakhiv (changes requested):

HStakhiv (approved):
Good!

TPolina (approved):
Good job!

PR #172, opened 2022-10-03
Solution

mariia-movchan:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file = command.split()[1]
```
mltdll, line 2: Generally, it's not a great idea to call a function twice where you can call it once. And here you can:
```python
command, file, new_file = command.split()

File: app/main.py [hunk 2]
```diff
@@ -1 +1,6 @@
...
+    new_file = command.split()[2]
+    if file != new_file:
```
mltdll, line 4: It's also generally better not to put function's main logic into an `if` block, even though in this case it doesn't make much difference. It would be better to do it like this:
```python
if file == new_file:
    return
...
```

You may also want to check if `command` == `cp`.

mltdll (changes requested):

mltdll (approved):

app/main.py
mariia-movchan: brilliant suggestion, thx
Fixed :)

app/main.py
mariia-movchan: fixed

TPolina (approved):
Well done!

PR #173, opened 2022-10-04
'Solution'

illiariaboy94:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command):
+    name_in, name_out = command.split()[1:]
+
+    if name_in == name_out:
```
HStakhiv, line 4: Make sure the correct command `cp` was entered

HStakhiv (changes requested):

app/main.py
illiariaboy94: Ok, done.

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command):
+    name_in, name_out = command.split()[1:]
+
+    if command != "cp" or name_in == name_out:
```
HStakhiv, line 4: It doesn't work. Because the `command` contains more words

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command):
```
HStakhiv, line 1: Add a type annotation

HStakhiv (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
```
HStakhiv, line 2: In my opinion, it is better to use a name for the variable that is different from the argument that the function accepts

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if command[1] == command[2]:
```
HStakhiv, line 3: Add a check `command[0] == "cp"`

HStakhiv (changes requested):

app/main.py
illiariaboy94: OK

app/main.py
illiariaboy94: I agree, did not pay attention

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    text = command.split()
```
HStakhiv, line 2: Choose a better name than just `text`

HStakhiv (changes requested):

app/main.py
Y-Havryliv, line 2: `text` is not really good name because it's not a text

Y-Havryliv (changes requested):

Y-Havryliv (approved):

PR #174, opened 2022-10-04
Solution

misha-doroshenko:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    existing_file = command.split(" ")[1]
```
Vasyl-Poremchuk, line 2: You can use list unpacking `copy_comand, existing_file, new_file = command.split()`. And then you can use `copy_command` for checking.

app/main.py
```diff
@@ -1 +1,14 @@
...
+    new_file = command.split(" ")[2]
+
+    if existing_file == new_file:
+        return
+
+    with (open(existing_file, "r") as f_read,
```
Vasyl-Poremchuk, line 8: You forgot to check if the command is `cp` (check `checklist` file in this module).

Vasyl-Poremchuk (changes requested):

Vasyl-Poremchuk (approved):
Well done!

TPolina (approved):
Well done!

PR #176, opened 2022-10-04
Solution

nailia-mamiedova:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str):
```
Abnormaltype, line 1: you should check the input command

Abnormaltype (changes requested):

Abnormaltype (changes requested):
you should check the input command

app/main.py
```diff
@@ -1 +1,9 @@
...
+    names = command.split()
+    cd_command = names[0]
+    file = names[1]
+    file_copy = names[2]
+    if file != file_copy and cd_command == "cd":
```
Abnormaltype, line 6: why cd?
image

Abnormaltype (changes requested):
also, please, fix flake8

PR #178, opened 2022-10-08
'Solution'

IvanGLS:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+import os
+
+
+def copy_file(command: str) -> None:
+    os.popen(command)
```
IvanRamyk, line 5: you should write your own implementation

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #182, opened 2022-10-16
Solution

Amirammi:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
```
DmytroSvirsa, line 2: Make sure you check that the `cp` input command is correct.

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #183, opened 2022-10-17
Solution

yessloss:


DmytroSvirsa (changes requested):
please read checklist(remove annessesary code, add check of command)

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_name = command.split()
+    if file_name[1] != file_name[2]:
```
DmytroSvirsa, line 3: Make sure you check that the `cp` input command is correct.

DmytroSvirsa (changes requested):

app/main.py
yessloss: in this condition I check if first file name = second file name, code stops

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #188, opened 2022-10-27
'Solution_#1'

Jumanje82:


Abnormaltype (changes requested):
you should check the input command as well

Abnormaltype (changes requested):
you should check if the input command is cp

Abnormaltype (approved):
Well done!

PR #192, opened 2022-10-30
Solution

the-andrew-007:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, filename_from, filename_to = command.split()
```
```suggestion
    operation, filename_from, filename_to = command.split()
```
check if the operation is equal to "cp"

IvanRamyk (changes requested):

IvanRamyk (approved):
Well done!

PR #197, opened 2022-11-09
solution

nazar-dmytrenko:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp = command.split()[0]
+
+    if cp != "cp" or command.split()[1] == command.split()[2]:
+        print("Does nothing")
```
DmytroSvirsa, line 5: but you don`t stop the execution, better to raise the custom Exception or stop the execution in another way

DmytroSvirsa (changes requested):

DmytroSvirsa (approved):

PR #199, opened 2022-11-12
Solution

Ida83:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_list = command.split()
+    if file_list[0] != "cp":
+        print("Incorrect command")
+    if file_list[1] == file_list[2]:
+        pass
```
Abnormaltype, line 6: for what od you need this?

Abnormaltype (changes requested):

app/main.py
Ida83: copy_file("cp file.txt file.txt")  # Does nothing
we do nothing if files have same names, but i think that all rest code should be after else:
am I correct?

Abnormaltype: it's not a solution for this

Abnormaltype: you can write 
```
if file_list[1] != file_list[2]:
        your code here
```

Abnormaltype (changes requested):

Abnormaltype (changes requested):

Abnormaltype (approved):
Well done!

PR #204, opened 2022-11-21
Solution

IvanMayor:


Greenmanxd:
is it ok that i had no "pytest"? Like i tryed tipe it but it didnt work

OleksandrPozden:
> is it ok that i had no "pytest"? Like i tryed tipe it but it didnt work

Yes, It is fine, because there are no tests in the ```test_main.py``` file

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splited_command = command.split()
```
OleksandrPozden, line 2: You can split it into three variables instead of holding them in the list.
Like this 
```command, file, new_file = command.split()```

File: app/main.py [hunk 2]
```diff
@@ -1 +1,11 @@
...
+
+    if splited_command[0] == "cp" and splited_command[1] != splited_command[2]:
+        with open(splited_command[1], "r") as \
+                old_file, open(splited_command[2], "w") as new_file:
+            read_file = old_file.read()
+            new_file.write(read_file)
+
+    elif command[1] == command[2]:
```
OleksandrPozden, line 10: check this condition at the beginning 
and use ```if``` instead of ```elif```

OleksandrPozden (changes requested):

app/main.py
Greenmanxd: Didnt know that thx

app/main.py
```diff
@@ -1 +1,15 @@
...
+    pass
+
+
+def copy_file(command: str) -> None:
+    command, file, new_file = command.split()
+
+    if file == new_file:
+        raise EqualNamesError("File names have not to be equal!!!")
+
+    elif command == "cp":
```
YuriiKindrat, line 11: u can use just if here instead of elif

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #211, opened 2022-11-28
'Copy_file_Solution'

MilManIT:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    first_name = command.split()[1]
+    second_name = command.split()[2]
+
```
OleksandrPozden, line 5: you can write in one row

OleksandrPozden (changes requested):

PR #212, opened 2022-11-30
Solution

Zetamag:


app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+from os.path import exists
+
+
+def copy_file(command: str) -> None:
+    tokens = command.split(" ")
```
YuriiKindrat, line 5: no need to write split(" "), it breaks string by the " " by default

YuriiKindrat (changes requested):

OleksandrPozden (approved):

PR #213, opened 2022-12-01
Solution

darynka-h:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    names = command.split()
```
```suggestion
    command, source, destination = command.split()
```
it's more clear to use unpacking here

IvanRamyk (changes requested):
wd

OleksandrPozden (approved):

PR #216, opened 2022-12-02
Solution

dmytro42:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(line: str) -> None:
+    f_names = line.split(" ")
```
OleksandrPozden, line 2:
```suggestion
    f_names = line.split()
```

OleksandrPozden: and unpack the variables, instead of using list

OleksandrPozden (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(line: str) -> None:
+    com, file1, file2 = line.split()
```
YuriiKindrat, line 2: use full names for variables - command not com
also it is better to use more informative variables not file1/file2 - new_file, old_file

YuriiKindrat (changes requested):

Abnormaltype (approved):

PR #226, opened 2022-12-13
Add files via upload

MilManIT:


main.py
```diff
@@ -0,0 +1,10 @@
+# write your code here
+def copy_file(command: str) -> None:
+
+    first_name, second_name = command.split()[1], command.split()[2]
```
IvanRamyk, line 4: It's better to create commands = command.split()
And use unpacking to get command and filenames

IvanRamyk (changes requested):

main.py
```diff
@@ -0,0 +1,10 @@
+# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split()
+    first_name, second_name = commands[1], commands[2]
```
OleksandrPozden, line 4:
```suggestion
     first_name, second_name = command.split()


```

OleksandrPozden (changes requested):

OleksandrPozden (approved):

PR #227, opened 2022-12-14
solution

kleotan901:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+import shutil
+
+
+def copy_file(string: str) -> None:
+    file_name = string.split(" ")[1]
```
IvanRamyk, line 5: use unpacking 
```
command, source, destination = string.split(" ")
```

IvanRamyk: it's more readable way to extract needed values

IvanRamyk (changes requested):

app/main.py
kleotan901: fixed

YuriiKindrat (approved):

PR #234, opened 2022-12-21
'Solution'

evgenijmartynuk07:


evgenijmartynuk07:
if use "cp" for Linux of "copy" for Win? with peopen I think will be correct 
if not, then why do we need a str with "cp" or "copy"?

Abnormaltype (changes requested):
you should implement your version of this functionality

Abnormaltype (changes requested):
use the checklist before pushing your solution

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split()
```
```suggestion
    command_type, source, destination = command.split()
```
it's more readable to use unpacking here

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #237, opened 2022-12-24
file_copy

VItalik1292:


app/main.py
```diff
@@ -1 +1,17 @@
...
+class CommandNotFound(Exception):
+    pass
+
+
+def copy_file(command: str) -> None:
+    elements_of_command = command.split()
+    try:
+        if elements_of_command[0] != "cp":
+            raise CommandNotFound
+    except CommandNotFound("You must enter a valid command"):
```
IvanRamyk, line 10: why do you need this `except ` if you still just `raise` the same error inside it

IvanRamyk: it seem that the whole try could be replaced with: 
```
if elements_of_command[0] != "cp":
            raise CommandNotFound("You must enter a valid command")
```

IvanRamyk (changes requested):

Vasyl-Smutok (approved):
well done👍

PR #240, opened 2022-12-27
Solution

xtsvirko:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_in_name, file_out_name = command.lstrip("cp").split()
```
```suggestion
    command_type, file_in_name, file_out_name = command.split()
```
more clear way is to use unpacking

IvanRamyk (changes requested):

IvanRamyk (approved):

PR #242, opened 2022-12-31
1st try solution

AnnaKabatova:


AnnaKabatova:
I don't know if it's gonna work because my "pytest" was empty and no tests ran

OleksandrPozden:
> I don't know if it's gonna work because my "pytest" was empty and no tests ran

You should try it out on your own.

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_ls = command.split()
+    file_old = command_ls[1]
+    file_new = command_ls[2]
```
OleksandrPozden, line 4: re-write in one line.
Use unpacking

OleksandrPozden: Also check whether command is "cp" nor something else

OleksandrPozden (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_ls = command.split()
+    file_old, file_new = command_ls[1], command_ls[2]
```
YuriiKindrat, line 3: u can write it like that command_cp, file_old, file_new = command.split() in one line)

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #260, opened 2023-01-21
'Solution'

timyr4iq:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command[3:].split(" ")
```
Vasyl-Smutok, line 2: better like this 
`cp, current_file, copied_file = command.split()`

Vasyl-Smutok (changes requested):
image

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import os
+
+
+def copy_file(command: str) -> None:
+    try:
+        cp, current_file, copied_file = command.split()
```
ArsenPidhoretskyi, line 6: cp here is bad naming because maybe someone enters another value.

app/main.py
```diff
@@ -1 +1,14 @@
...
+    except ValueError:
+        raise ValueError("Command is incorrect")
+    if cp != "cp" or os.path.exists(current_file) is False:
```
```suggestion
    if cp != "cp" or not os.path.exists(current_file):
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+        raise ValueError("Command is incorrect")
+    if current_file != copied_file:
+        with open(current_file,
+                  "r") as file_in, open(copied_file, "w") as file_out:
+            file_out.write(file_in.read())
```
ArsenPidhoretskyi, line 14:
```suggestion
        with (
            open(current_file, "r") as file_in,
            open(copied_file, "w") as file_out
        ):
            file_out.write(file_in.read())

```

app/main.py
ArsenPidhoretskyi: Also you need to validate command, if here more than 3 splitted values: cp file1.txt file2.txt file2.txt

ArsenPidhoretskyi (changes requested):

app/main.py
timyr4iq: that's why I use the try command. If there are more than three values, a ValueError will appear

ArsenPidhoretskyi: Oh sorry, my bad

ArsenPidhoretskyi: Just rename cp

timyr4iq: but why is it worth renaming cp, because in this particular case it should be the cp command and then I even check it

timyr4iq: the previous reviewer even advised me to write this way

ArsenPidhoretskyi: You can say that is 'cp' after checking, before check it just unknown command)
Better to use command or input_command.

ArsenPidhoretskyi (approved):
Good!

PR #264, opened 2023-01-25
Solution

olenaliuby:


app/main.py
```diff
@@ -1 +1,15 @@
 # write your code here
+def copy_file(command: str) -> None:
+    try:
+        command_cp, file_name, file_copy_name = command.split()
```
ArsenPidhoretskyi, line 4: Please rename variables.
command_cp -> In this step we can not say that it is cp_command. Just command or input_command
file_name -> Why need file_name? input_file_name or source_file_name
file_copy_name -> Why need file_copy_name? output_file_name or destination_file_name

app/main.py
```diff
@@ -1 +1,15 @@
...
+    except ValueError:
+        raise ValueError(
+            "Check if the command has the format: cp file.txt file_copy.txt"
+        )
+    if command_cp != "cp":
+        raise ValueError("The command must start with 'cp'")
+    if file_name == file_copy_name:
+        raise ValueError("The file copy must have different name")
+    with open(file_name, "r") as file_in, \
+            open(file_copy_name, "w") as file_out:
```
ArsenPidhoretskyi, line 14:
```suggestion
    with (
        open(file_name, "r") as file_in,
        open(file_copy_name, "w") as file_out
    ):
```

ArsenPidhoretskyi (changes requested):

app/main.py
olenaliuby: This definitely looks better. Thank you for a hint!

ArsenPidhoretskyi (approved):
In task we do not say to raise an error if command != 'cp' and if file names is the same.
But for me it looks good!

PR #267, opened 2023-01-28
solution

frameq:


Vasyl-Smutok (changes requested):
image

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import shutil
+
+
+class CommandTypeError(Exception):
+    pass
+
+
+def copy_file(command: str) -> None:
+    command, input_file, output_file = command.split(" ")
```
ArsenPidhoretskyi, line 9: I can get an error in those cases. Validate this
```
command == "mv file1.txt"
command == "mv file1.txt file2.txt file3.txt"
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+import shutil
+
+
+class CommandError(Exception):
+    pass
+
+
+def copy_file(command: str) -> None:
+    try:
+        command, input_file, output_file = command.split(" ")
```
olena-hul, line 10: Check the length of command, it should be exactly 3 arguments, command, input file and output file. And there is no need to pass space to .split(" ").
.split() with no args will work the same because space is default arg here

app/main.py
```diff
@@ -1 +1,15 @@
...
+    except ValueError:
+        raise CommandError("Check command arguments")
+    if command != "cp":
+        raise CommandError("Check input command")
+    shutil.copyfile(input_file, output_file)
```
olena-hul, line 15: You should implement your custom solution, and not to use libraries

olena-hul (changes requested):

Vasyl-Smutok (approved):
Well Done 👍

PR #268, opened 2023-01-29
Solution

Phoenix-Erazer:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, original_f, copy_f = command.split()
```
olena-hul, line 2: Validate command length, it should be exaclty 3 parts

olena-hul: And use better naming

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
+        return print("Command length is not equal to 3")
+
+    action, original_file, new_file = command.split()
+    if action != "cp" or original_file == new_file:
```
katevasilenko, line 6: I disagree about second part `original_file == new_file ` as your code should do nothing in this particular situation but you give a message about incorrect command. It could lead to misunderstanding

katevasilenko (changes requested):

Abnormaltype (approved):
Well done!

PR #271, opened 2023-02-02
'Solution'

Vitalii2906:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    command = command.split(" ")
```
katevasilenko, line 3: You can unpack command to three variables in one line. Also don’t use space in split as it use space by default
```suggestion
    command, old_file, new_file = command.split()
```

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if command[1] == command[2]:
+        print("Does nothing")
```
katevasilenko, line 5: You don’t need to print this string)

katevasilenko (changes requested):
Also read checklist and add changes to your code

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    command, old_file, new_file = command.split(" ")
```
ArsenPidhoretskyi, line 3: Validate command
You code will not work in this cases:
command = 
""
"mv"
"mv file1.txt"
"mv file1.txt file2.txt file3.txt"

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    if command.count(" ") > 2 or command.count(" ") < 2:
+        return "wrong command"
```
Vasyl-Smutok, line 4:
```suggestion
    if command.count(" ") != 2
        return "wrong command"
```


but your check  doesn't work when I type `сp cp cp`

Vasyl-Smutok (changes requested):
image

Vasyl-Smutok (changes requested):
Please check comments

app/main.py
ArsenPidhoretskyi: Just check as:
`if len(command.split()) != 3`

app/main.py
```diff
@@ -1 +1,23 @@
...
+    # separate the command and file names
+    command, old_file, new_file = command.split(" ")
+
+    # check that the file names do not match
+    if old_file == new_file:
+        return
+
+    # check whether the file names and commands are specified correctly
+    if old_file.endswith(".txt") \
+            and new_file.endswith(".txt") \
```
ArsenPidhoretskyi, line 17: As for me, No need to check for .txt

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+def copy_file(command: str) -> None:
+    """Copies the contents of one file to another"""
+
+    # we check the command structure
```
IvanRamyk, line 4: Please, remove the redundant comment

IvanRamyk: Don't leave comments in the final version of the code
Code should self-descriptive enough

IvanRamyk: please, fix everywhere

IvanRamyk (changes requested):

Grayder0152 (approved):

PR #274, opened 2023-02-05
Solution

avkpol:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> str:
+    file_1, file_2 = command.split()[1], command.split()[2]
```
ArsenPidhoretskyi, line 5: Okay if we will have 100 arguments solution will be the same?
1. https://github.com/avkpol/py-copy-file/blob/develop/checklist.md#code-efficiency
2. Validate that command have 3 arguments
3. Use correct variables names

app/main.py
```diff
@@ -1 +1,12 @@
...
+    with open(file_1, "r") as file_in, open(file_2, "w") as file_out:
+        if file_in == file_out:
+            pass
```
ArsenPidhoretskyi, line 8: This is need to check before open because you do additional actions

app/main.py
```diff
@@ -1 +1,12 @@
...
+        return shutil.copyfile(file_1, file_2)
```
ArsenPidhoretskyi, line 9: No need shututil use standard libraries

ArsenPidhoretskyi (changes requested):

app/main.py
avkpol: as far as I know, using side libraries is the proper pythonic way to write code, why not? Why use tons of code if it can be written in one line?

app/main.py
avkpol: sorry, I didn't get what you mean.

app/main.py
avkpol: 1. I stuck with the task's readme, cp command accepts only two arguments, doesn't it? Not 100.  
2. Should I  validate that the command has the "cp" keyword of the command itself and two more filename arguments?

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
+        raise TypeError(f"the command should have 3 arguments,"
```
IvanRamyk, line 3: `ValueError` is more suitable as for me

app/main.py
```diff
@@ -1 +1,15 @@
...
+                        f" got {len(command.split())} instead")
+    if "cp" not in command.split():
+        raise TypeError("the command should have 'cp' keyword, "
+                        "got '%s' instead" % command.split()[0])
+    source_f, result_f = command.split()[1], \
+        command.split()[2]
+    if source_f == result_f:
+        raise TypeError(f"{source_f} and {result_f} are the same file")
+    else:
```
IvanRamyk, line 12: Please, remove the redundant else

IvanRamyk (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
+        raise ValueError(f"the command should have 3 arguments,"
+                         f" got {len(command.split())} instead")
+    if "cp" not in command.split():
+        raise TypeError("the command should have 'cp' keyword, "
+                        "got '%s' instead" % command.split()[0])
+    source_f, result_f = command.split()[1], \
+        command.split()[2]
```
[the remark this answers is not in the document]
avkpol: I don't know beforehand what exact string I will get, so I split twice to obtain the "slash" delimited part and my part with the command itself and the first filename. I couldn't imagine another way to do that. My code is wrong?

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, source_f, result_f = command.split()
+    if len(command.split()) != 3:
```
Grayder0152, line 3: Check this before 2 line

Grayder0152 (changes requested):

IvanRamyk (approved):

PR #279, opened 2023-02-08
Solution

leetwinoff:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, old_name, new_name = command.split()
```
Grayder0152, line 2: What if comamnd is empty? Pls, check len of command

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if len(cp) != 0:
+        if old_name != new_name:
+            with open(old_name, "r") as file_out,\
+                    open(new_name, "w") as file_in:
```
Vasyl-Smutok, line 6: better to use `()` than `\`

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, old_name, new_name = command.split()
+    if len(cp) != 0:
```
Vasyl-Smutok, line 3: bad check

image

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, old_name, new_name = command.split()
+    if old_name != new_name and cp == "cp" and len(cp) != 0:
+        with open(old_name, "r") as file_in:
+            with open(new_name, "w") as file_out:
```
Bohdan-Salamakha, line 5:
```
        with (open(old_name, "r") as file_in,
                 open(new_name, "w") as file_out):
```

Bohdan-Salamakha (changes requested):
v1

app/main.py
leetwinoff: What do you mean?

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command) != 0:
+        cp, old_name, new_name = command.split()
+        if old_name != new_name and cp == "cp":
+            with open(old_name, "r") as file_in:
+                with open(new_name, "w") as file_out:
```
Vasyl-Smutok, line 6:
```suggestion
            with (
                open(old_name, "r") as file_in,
                open(new_name, "w") as file_out
            ):
```

Vasyl-Smutok (changes requested):

app/main.py
Vasyl-Smutok: I think Bohdan had this in mind)))
```
with (
    open(old_name, "r") as file_in,
    open(new_name, "w") as file_out
):
```

leetwinoff: when I make this way I have invalid syntax error. If I use slash everything ok. That's why I changed with 2 WITH OPEN

app/main.py
leetwinoff: when I make this way I have invalid syntax error. If I use slash everything ok. 
That's why I changed with 2 WITH OPEN

Grayder0152: Pls, use Python 3.10 to avoid this issue in a future

Grayder0152 (approved):

PR #282, opened 2023-02-09
Solution

oleksandr-lysanets:


oleksandr-lysanets:
I think I found a better solution

Bohdan-Salamakha (approved):

PR #283, opened 2023-02-11
solution

MaxiTDI:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, file_name, copy_file_name = command.split()
+    if file_name == copy_file_name or cmd != "cp":
+        return
```
Vasyl-Smutok, line 4: I think better to use `raise` some `Exception` 
and add descriptions

Vasyl-Smutok (changes requested):

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    # Split command line into three parts
+    cmd, file_name, copy_file_name = command.split()
```
ArsenPidhoretskyi, line 4: If command.split() will have 2 or 4 elements your code will broke
Validate this

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #285, opened 2023-02-12
1 commit

AndereLion:


manjustice (approved):
GJ!

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file, file_copy = command.split(" ")
```
YuriiKindrat, line 2: cp is not descriptive name for variable

YuriiKindrat: Also no need to pass " " in split method it use spaces by default

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if file != file_copy and cp == "cp":
+        with open(file, "r") as file_in, open(file_copy, "w") as file_out:
+            data = file_in.read()
+            file_out.write(data)
```
YuriiKindrat, line 6: no need to create variable here, u can call method read inside method write

YuriiKindrat (changes requested):

app/main.py
AndereLion: Fixed

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp_command, file, file_copy = command.split()
+    if file != file_copy and cp_command == "cp":
+        with open(file, "r") as file_in, open(file_copy, "w") as file_out:
+            data = file_in.read()
+            file_out.write(data)
```
Serhii-Leonenko, line 6: remove redundant variable `data` as Yurii said

Serhii-Leonenko (changes requested):

app/main.py
AndereLion: Fixed

Serhii-Leonenko (approved):
Well done!

PR #288, opened 2023-02-13
Solution

SafonovVladimir:
add my solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,9 @@
-# write your code here
+def copy_file(command_line: str) -> None:
+    command, old_file, new_file = command_line.split(" ")
+    content = ""
+    if command == "cp":
+        with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+            # collect all text from old file to variable content
+            content += "".join(file_in.readlines())
+            file_out.writelines(content)
+            print(f"File {old_file} has been copied to the file {new_file}")


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command_line: str) -> None:
+    command, old_file, new_file = command_line.split(" ")
```
bihunva, line 2: "split" method, splits the string by spaces by default

bihunva (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
...
+    content = ""
+    if command == "cp":
```
bihunva, line 4: it must do nothing in case the user trying to copy file to file with the same name

SafonovVladimir: Fixed

app/main.py
SafonovVladimir: Ah, U r right.
Fixed

MaybeTI (approved):
GJ!

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command_line: str) -> None:
+    command, old_file, new_file = command_line.split()
+
+    if command == "cp" and old_file != new_file:
+        with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+            file_out.write(file_in.read())
+            print(f"File {old_file} has been copied to the file {new_file}")
```
YuriiKindrat, line 7: this is a requirement from checklist to the task, fix this please
image

YuriiKindrat (changes requested):

app/main.py
SafonovVladimir: Done

YuriiKindrat (approved):

PR #290, opened 2023-02-13
Solution

MaybeTI:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    ls_file = command.split()
+    if ls_file[1] != ls_file[2] and ls_file[0] == "cp":
```
SafonovVladimir, line 3: U can unpacking in 3 variables in one line
Ex. command, old_file, new_file = command.split() 
where ls_file[0] == command ...

SafonovVladimir (changes requested):

SafonovVladimir (approved):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    func, file_1, directory = command.split()
```
YuriiKindrat, line 2: bad name for variables, use more descriptive names for naming, f.e. func stores command name, so u can name it like "command" or smth else

YuriiKindrat (changes requested):

app/main.py
SafonovVladimir: Fixed

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file, directory = command.split()
```
Serhii-Leonenko, line 2: `directory` also is a bad name for the new file, change it

Serhii-Leonenko (changes requested):

amber-marichi (approved):
looks nice

PR #291, opened 2023-02-14
Solution

xtina999:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    data = command.split()
```
ArsenPidhoretskyi, line 2: Validate command for three elements. Also you can use as:
command, source_filepath, destination_filepath = command.split()

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,5 +1,6 @@
 def copy_file(command: str) -> None:
-    data = command.split()
-    if data[1] != data[2] and data[0] == "cp":
-        with open(data[1], "r") as file_in, open(data[2], "w") as file_out:
+    command, source_filepath, destination_filepath = command.split()
+    if source_filepath != destination_filepath and command == "cp":
+        with open(source_filepath, "r") as file_in, \
```
vasylhnatiuk, line 4: would be better to add brackets       
```
  with( 
open(source_filepath, "r") as file_in, 
open(destination_filepath, "w") as file_out
      ):
```

vasylhnatiuk (approved):

PR #292, opened 2023-02-14
solution

OleksandrGrk:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    data = command.split()
```
katevasilenko, line 2: You can unpack all three items directly on this line

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if data[0] == "cp":
+        file_1 = data[1]
+        file_2 = data[2]
```
katevasilenko, line 5: It is not really good names for variables. Maybe something like `old_file` and `new_file`

katevasilenko (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    content = command.split()
+    command, old_file, new_file = content[0], content[1], content[2]
```
Grayder0152, line 3: You can do it un one line, but before you need to check that `command` length == 3

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if command == "cp":
+        with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+            if old_file != new_file:
```
Grayder0152, line 6: lest check it before open files

Grayder0152 (changes requested):

app/main.py
OleksandrGrk: Done

app/main.py
OleksandrGrk: But command is a string, length of command always > 3

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    content = command.split()
+    command, old_file, new_file = content[0], content[1], content[2]
```
ArsenPidhoretskyi, line 3: If splited command length less that 3 elements your program will not work
You must validate this

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    content = command.split()
+    if len(content) == 3:
+        command, old_file, new_file = content[0], content[1], content[2]
+        if command == "cp" and old_file != new_file:
+            with open(old_file, "r") as file_in, \
+                    open(new_file, "w") as file_out:
```
```suggestion
            with (
                open(old_file, "r") as file_in,
                open(new_file, "w") as file_out
            ):
```

ArsenPidhoretskyi (changes requested):

app/main.py
OleksandrGrk: thanks! Looks better

ArsenPidhoretskyi (approved):
Well done!

PR #293, opened 2023-02-14
Develop

AlimFreight:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
katevasilenko, line 2: You can unpack all three values directly on this line

katevasilenko (approved):

PR #295, opened 2023-02-15
1

DavidGlusch:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.split()[1] == command.split()[2]:
+        return None
```
ZabFTFT, line 3: As I realized we also need to check: "Make sure you check that the `cp` input command is correct". In your case we need to check command.split()[0] == "cp". But I am not sure, so .. =)

ZabFTFT (approved):

app/main.py
Serhii-Leonenko, line 3: no need som many split, you can unpack values from the command with one split().
 `variable, variable, variable = command.split()`

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, origin, copy = command.split()
+
+    if origin == copy or cp == "cp":
+        return None
```
YuriiKindrat, line 5: wrong condition here, we should use cp command
also no need to use return None here, u can just use backward condition for code below

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, origin, copy = command.split()
```
YuriiKindrat, line 2: bad names for variables, use more descritptive names, f.e. not cp, but command

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #296, opened 2023-02-15
Solution

ZabFTFT:


NEkropo1 (changes requested):
All good, but I'm pretty sure you don't need `if name block` 'cause it was only for testing purpose.

NEkropo1 (approved):
seems good )

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    input_command, file_name_to_copy, new_file_name = command.split()
+    if input_command != "cp" or file_name_to_copy == new_file_name:
+        return
```
YuriiKindrat, line 4: no need to use blank return here, use backward condition instead

app/main.py
```diff
@@ -1 +1,9 @@
...
+
+    with (open(file_name_to_copy, "r") as file_in,
+            open(new_file_name, "w") as file_out):
+        read_content = file_in.read()
+        file_out.write(read_content)
```
YuriiKindrat, line 9: no need to create variable here, u can call method read inside method write

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #297, opened 2023-02-15
Solution, but need suggestions

NEkropo1:


ZabFTFT (approved):
Well done!

kuznetsb (changes requested):
Check the need of close() function usage

app/main.py
```diff
@@ -1 +1,13 @@
...
+        raise AttributeError(f"Name of copy command "
+                             f"should be 'cp', not {copy}")
+
+    if file_name == new_file_name:
+        return
+
+    with open(file_name, "r") as file_in, open(new_file_name, "w") as file_out:
+        file_out.write(file_in.read())
+        file_in.close()
+        file_out.close()
```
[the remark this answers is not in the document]
NEkropo1: Yup I understand that we work with file managers here and they automatically close files. But I seen in some companies that after working even with content managers, nevertheless, they put file,close() after working with file for excessively data saving, like for SEU, and I just was wandering either my code is ok or or we can totally rely on DM.

kuznetsb (approved):
Let's see what mentor say))

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    copy, file_name, new_file_name = command.split()
```
YuriiKindrat, line 2: bad name for command variable, use more descriptive name, f.e. command

app/main.py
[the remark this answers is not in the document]
YuriiKindrat: @kuznetsb is right, no need to use close() method inside CM

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if copy != "cp":
+        raise AttributeError(f"Name of copy command "
```
YuriiKindrat, line 4: are u sure about this [exception](https://www.geeksforgeeks.org/python-attributeerror/#:~:text=AttributeError%20can%20be%20defined%20as%20an%20error%20that%20is%20raised%20when%20an%20attribute%20reference%20or%20assignment%20fails.%C2%A0) ?

YuriiKindrat (changes requested):

app/main.py
NEkropo1: Can't get the idea. I mean we get some command, which we split on command + name + name,which name should be good?

app/main.py
NEkropo1: AttributeError can be defined as an error that is raised when an attribute reference or assignment fails. 
For example, if we take a variable x we are assigned a value of 10. In this process suppose we want to append another value to that variable. It’s not possible. Because the variable is an integer type it does not support the append method

Seems more like ValueError)), but  I didn't define it with exact definition, 'cause, if it's command which is assigned to some method inside class, it seems to me more logical, that we cant attach some other command 'cause it shouldn't support this method. Just wanted to get some answers. Pretty sure it's ValueError)

Serhii-Leonenko (approved):
Good job!

PR #300, opened 2023-02-16
Solution

WDemidenko:


NEkropo1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,10 @@
 def copy_file(command: str) -> None:
-    command = command.split(" ")
-    if len(command) != 3 or command[0] != "cd" or command[1] == command[2]:
+    try:
+        command, file_original, file_copy = command.split(" ")
```
NEkropo1, line 3:
```suggestion
        command, file_original, file_copy = command.split()
```

NEkropo1 (changes requested):

NEkropo1 (approved):
I still recommend you to use all you don't need to test outside try-except block, but it's on you to decide.
P.S. `if file_original == file_copy: print("Wrong command!")` seems not very logical to me )

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        command, file_original, file_copy = command.split()
+        if command != "cd":
+            raise ValueError
+    except ValueError:
+        print("Wrong command!")
+        return None
```
Serhii-Leonenko, line 8: For what do you raise and then handle the error?

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if file_original == file_copy:
+        print("Files must have different names.")
+        return None
```
Serhii-Leonenko, line 11: Use return  without None

Serhii-Leonenko (changes requested):
And no need to use so many `return` here, you can do the same without it at all

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_original, file_copy = command.split()
+    if command != "cd":
+        raise ValueError("Wrong command!")
+    if file_original == file_copy:
+        return
+    with open(file_original, "r") as original, open(file_copy, "w") as copy:
+        copy.write(original.read())
```
YuriiKindrat, line 8: no need to use blank return here, u can use backward condition for this code

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #302, opened 2023-02-16
Solution

Yevheniy-Lototskiy:


ZabFTFT (approved):
Nice startwith usage, like it! Also you can make one line instead of 2 with unpacking but its up to you. Well done!

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_org = command.split()[1]
+    file_copy = command.split()[2]
```
YuriiKindrat, line 3: make ur code dry, u can unpack this to variables like this `command, file_org, file_copy = command.split()`

YuriiKindrat (changes requested):

PR #305, opened 2023-02-17
Solution

oksana-feshchenko:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, first, second = command.split(" ")
```
Serhii-Leonenko, line 2: give more descriptive names for variables, also `split()` already has a default value `" "`  no need to write it

Serhii-Leonenko (changes requested):

PR #306, opened 2023-02-17
Solution

Maytreyaya:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str):
+    res = command.split(" ")
+    if res[0] == "cp":
+        with open(res[1], "r") as source:
+            data = source.read()
+
+        with open(res[2], "w") as target:
+            target.write(data)
```
Morty67, line 8: you can use like in example  

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str):
+    res = command.split(" ")
```
Morty67, line 2: i think it's not good name for variable

Morty67 (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    result = command.split(" ")
```
YuriiKindrat, line 2: no need to use space as arg in split method, it uses space by default

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if result[0] == "cp":
```
YuriiKindrat, line 3: what if files will have the same name ?
image

app/main.py
YuriiKindrat: also no need to take values by index, u can use split method to unpack splitted string to variables

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file, new_file = command.split("")
```
Serhii-Leonenko, line 2: As Yurii said, "no need to use space as arg in split method, it uses space by default"

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if command == "cp":
+        with open(file, "r") as source, open(new_file, "w") as target:
+            data = source.read()
+            target.write(data)
+    if file == new_file:
+        return
```
Serhii-Leonenko, line 8: You can use one `if` condition with logical operator

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1,8 +1,6 @@
 def copy_file(command: str) -> None:
-    command, file, new_file = command.split("")
-    if command == "cp":
+    command, file, new_file = command.split()
+    if command == "cp" and file != new_file:
         with open(file, "r") as source, open(new_file, "w") as target:
             data = source.read()
             target.write(data)
```
Serhii-Leonenko, line 6: Sorry, I didn't say it before, but you can remove the redundant variable in this case and make it in one line

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #307, opened 2023-02-17
'Solution'

Morty67:


usrofgh (approved):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
+    if command.startswith("cp") and split_command[1] != split_command[2]:
+        with open(split_command[1], "r") as file_in, open(
+            split_command[2], "w"
```
YuriiKindrat, line 5: no need to take values by index, u can use split method to unpack splitted string to variables

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #309, opened 2023-02-17
solution

usrofgh:


app/main.py
```diff
@@ -1 +1,11 @@
...
+    try:
+        cd, old_file, new_file = command.split()
+    except ValueError:
+        raise ValueError("Write a command correctly. "
+                         "Example: cp file1.txt file2.txt")
+    if old_file == new_file:
+        return
+
+    with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+        file_out.write(file_in.read())
```
ZabFTFT, line 11: You can combine condition and with block here. Definitely mentor Yurii would reject this.I don't know why but we shouldn't make blank return without huge reason. Got several rejects cause of it.

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        cd, old_file, new_file = command.split()
+    except ValueError:
+        raise ValueError("Write a command correctly. "
+                         "Example: cp file1.txt file2.txt")
```
ZabFTFT, line 6: We have some condition in checklist: "Make sure you check that the `cp` input command is correct.". In your case you are raising exception when command is int for example =) Its good that you mentioned correct format but we can type any command here, like "mv file1.txt file2.txt" and it will work with this code.Correct me if I am wrong.

vladzabolotnyi (changes requested):

vladzabolotnyi (approved):
Well done!

SergiiMachulin (approved):

SergiiMachulin:
LGTM, but why do you need the last line(call) in your code?)

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        cp, old_file, new_file = command.split()
+        if cp != "cp" or old_file == new_file:
+            raise ValueError
+    except ValueError:
+        raise ValueError("Write a command correctly. "
```
YuriiKindrat, line 7: why do u need to use ValueError twice? u can catch a several exceptions in ur except block

app/main.py
```diff
@@ -1 +1,14 @@
...
+                         "Example: cp file1.txt file2.txt")
+
+    with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+        file_out.write(file_in.read())
+
+
+copy_file("cp file1.txt file2.txt")
```
YuriiKindrat, line 14: remove unnecessary call of a function

YuriiKindrat (changes requested):
also remove unnecessary files from ur PR

app/main.py
usrofgh: It seems to me if I delete raise in if condition, then I won't get raise error when ```cp``` var won't be equal to "cp", thus my command will be able to execute for example with the "move" command

> why do u need to use ValueError twice? u can catch a several exceptions in ur except block

YuriiKindrat (approved):

PR #310, opened 2023-02-18
Develop

yura-hudzovskyi:


DmytryTong (commented):
You should check: if name of a source file is not the same as name of a new file.

PR #311, opened 2023-02-18
Develop

yura-hudzovskyi:


inrock1 (changes requested):
One of the tasks was:
The function must do nothing in case the user is trying to copy file to file with the same name.

inrock1 (approved):
By the way, you can combine several "if" into one "if" with "and"

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_name = command.split()
+
+    if file_name[1] != file_name[2]:
+        if file_name[0] == "cp":
```
YuriiKindrat, line 5: no need to take values by index, u can use split method to unpack splitted string to variables
arg1, arg2, arg3 = ur_string.split()

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
...
+    command, old_file, new_file = command.split()
+
+    if old_file != new_file and command == "cp":
+        with open(
+                old_file,
+                "r"
+        ) as file, open(
+            new_file,
+            "w"
+        ) as file_copy:
```
Serhii-Leonenko, line 11: It's really hard to read in this case, try to do it with brackets in two lines

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #313, opened 2023-02-18
Solution

SergiiMachulin:


ZhAlexR (approved):
Good job!

RuslanLopatin (approved):

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
...
+            open(file_in_name, "r") as file_in,
+            open(file_out_name, "w") as file_out
+        ):
+            file_out.write(file_in.read())
+    except ValueError:
+        print(
+            "Your command must consist of: keyword 'cp',"
+            " file name to copy and new file name, "
+            "separated by spaces"
+        )
```
Serhii-Leonenko, line 16: No need to raise the exception, and then handle it and print the message. Raise the error or print the message or move logic into condition without everything else.

Serhii-Leonenko (changes requested):

app/main.py
Serhii-Leonenko: image

Serhii-Leonenko (approved):
Well done!

PR #314, opened 2023-02-18
Solution DonchykD

DmytryTong:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    temp = command.split(" ")
+    if temp[0] != "cp":
+        return 0
+    if temp[1] == temp[2]:
```
BohdanLazaryshyn, line 5: you can write it in one line

BohdanLazaryshyn (changes requested):

app/main.py
DmytryTong: Yep, I guess I can - fixed)

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    temp = command.split(" ")
```
YuriiKindrat, line 2: No need to use space as arg for split method, it use space by default

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if temp[0] != "cp" or temp[1] == temp[2]:
```
YuriiKindrat, line 3: u can unpack this variables using split like this: `arg1, arg2, arg3 = some_string.split()`
also no need to return 0 here, u can just use backward condition

YuriiKindrat (changes requested):

app/main.py
DmytryTong: Ok, fixed)

app/main.py
DmytryTong: Fixed)

YuriiKindrat (approved):

PR #315, opened 2023-02-18
'Solution'

inrock1:


app/file.txt
```diff
@@ -0,0 +1,6 @@
+weqf
```
NazarHladaniuk, line 1: you commit and push test file?

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3 or command[:3] != "cp ":
+        return 0
+    name_f_in = command.split()[1]
+    name_f_out = command.split()[2]
```
NazarHladaniuk, line 5: U can use tuple unpacking again to assign the values of name_f_in and name_f_out without calling split() twice. 
If you want to declare these variables like this, it will help to optimize the code

NazarHladaniuk (changes requested):

app/file.txt
inrock1: no. I committed app/main.py
file.txt was added automatically.

app/main.py
inrock1: Yes. It is possible. But this decision is the same as mine.
No matter, look for a new tuple or given initial string.
And in my case, I don't occupy the memory of the new variable.

NazarHladaniuk: Ok, but using variable, variable, variable = command.split() and command.startswith can make your code better

NazarHladaniuk (approved):

app/file.txt
YuriiKindrat: remove unnecessary file from ur PR, please

app/main.py
YuriiKindrat: @NazarHladaniuk is right, no need to call split method for 3 times, also it is hard to read ur code
for better code style u can use split for unpack ur splitted string to variables f.e.
also try to avoid using return 0 in ur code, u can just make a one backward condition to copy ur file

YuriiKindrat (changes requested):

app/main.py
inrock1: `arg1, arg2, arg3 = ur_string.split()`
Now I understand what @NazarHladaniuk offered me, yesterday word 'tuple' confused me.

All comments are taken into account.

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    short_command, name_f_in, name_f_out = command.split()
```
YuriiKindrat, line 2: great, but use full names for ur variables

inrock1: corrected

Serhii-Leonenko (approved):
Well done!

PR #316, opened 2023-02-18
'Solution'

IvanGuculyak:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    arguments = [word for word in command.split()]
```
Vasyl-Smutok, line 2:
```suggestion
    cp_command, current_file, future_file = command.split()
```

Vasyl-Smutok (changes requested):
image

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp_command, current_file, future_file = command.split()
```
ArsenPidhoretskyi, line 2: Your code will broke in next cases:
command = 
"cp"
"cp file1.txt"
"cp file1.txt file2.txt file3.txt"

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if current_file == future_file and cp_command == "cp":
```
ArsenPidhoretskyi, line 3: if current_file == future_file but command will not equal to cp -> we will do copy
Strange logic

ArsenPidhoretskyi (changes requested):

katevasilenko (approved):
Good job!

PR #317, opened 2023-02-18
Solution

ZhAlexR:


ZabFTFT (approved):
Well done to use extended library ;)

Anatolii-Poznyak (approved):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+from shutil import copyfile
+
+
+def copy_file(command: str) -> None:
+    execution_command, curr_file, new_file = command.split(" ")
```
YuriiKindrat, line 5: no need to pass space as arg to a split method, it uses space by default

YuriiKindrat: also use full names for variable, current - not curr

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #318, opened 2023-02-19
'Solution'

AnnaBoichenko:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    temp = command.split(" ")
+    old_file = temp[1]
+    new_file = temp[2]
+    if old_file == new_file:
+        exit()
+    with open(old_file, "r") as old:
+        with open(new_file, "w") as new:
```
oksana0410, line 8: you can write 7 and 8 in 1 line

oksana0410 (approved):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    temp = command.split(" ")
+    old_file = temp[1]
+    new_file = temp[2]
+    if old_file == new_file:
+        exit()
```
YuriiKindrat, line 6: 1. no need to use space as arg for split method, it use space by default
2. u can unpack command and files using only one line
3. no need to use exit method here, u can just make a backward condition
4. u should check is "cp" command passed
image

app/main.py
YuriiKindrat: @oksana0410 is right, fix this please
image

YuriiKindrat: also new and old is a bad names for variables, use more descritptive names for naming

app/main.py
AnnaBoichenko: Could you please clarify your point about unpacking commands? 
I think, using old_file and new_file for variables makes code more descriptive. 
And split method returns a list, so I decided to store it in variable temp for efficiency reasons, just to not split string each time we call the method

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.startswith("cp"):
+        temp = command.split()
+        old_file, new_file = temp[1], temp[2]
+        if temp[1] != temp[2]:
```
Serhii-Leonenko, line 5: You can unpack values from the tuple into variables like this `command, old_file, new_file = command.split()`
And then use these variables without any indexes)
Also, you can use just one if condition with logical operator

app/main.py
```diff
@@ -1 +1,9 @@
...
+            with open(old_file, "r") as file_in, open(
+                new_file, "w"
+            ) as file_out:
```
Serhii-Leonenko, line 8: You can increase readability in this case with brackets:

`(with open(old_file, "r") as file_in,`
`with open(new_file, "w") as file_out):`

Serhii-Leonenko (changes requested):

app/main.py
AnnaBoichenko: my python 3.7 doesn`t support it 
image

Serhii-Leonenko: ((

Serhii-Leonenko (approved):
Good job!

PR #320, opened 2023-02-19
'Solution'

RuslanLopatin:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def copy_file(command: str) -> None:
+    components = command.split()
+    if len(components) != 3:
+        raise ValueError("Invalid command format. "
+                         "Usage: cp <source_file> <destination_file>")
```
ada-krav, line 8: You should also check if the first(zeroest?) component is "cp" and not something else

app/main.py
```diff
@@ -1 +1,20 @@
...
+    destination_file = components[2]
+
+    if source_file == destination_file:
+        return
+
+    if not os.path.exists(source_file):
+        raise ValueError(f"Source file '{source_file}' does not exist")
+
+    with open(source_file, "rb") as fsrc, open(destination_file, "wb") as fdst:
+        fdst.write(fsrc.read())
```
ada-krav, line 20: fsrc and fdst might not be the best names

ada-krav (changes requested):

ada-krav (approved):

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import os
+
+
+def copy_file(command: str) -> None:
+    components = command.split()
+    if len(components) != 3 or components[0] != "cp":
```
YuriiKindrat, line 6: no need to use index here, u can use split method to unpack splitted string to varaiables, like this
`var1, var2, var3 = ur_string.split()`

app/main.py
```diff
@@ -1 +1,22 @@
...
+        raise ValueError("Invalid command format. "
+                         "Usage: cp <source_file> <destination_file>")
+
+    source_file = components[1]
+    destination_file = components[2]
+
+    if source_file == destination_file:
+        return
```
YuriiKindrat, line 14: no need to use blank return here, make a backward condition for code below

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    if not os.path.exists(source_file):
+        raise ValueError(f"Source file '{source_file}' does not exist")
+
+    with open(source_file, "rb") as sourcefile, \
+            open(destination_file, "wb") as destin_file:
```
YuriiKindrat, line 20: try to avoid using " \ " in ur code, u can split this line by bracket

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: also fix ur flake8 errors

app/main.py
```diff
@@ -1 +1,18 @@
...
+def copy_file(command: str) -> None:
+    command, source_file, destination_file = command.split()
+
+    if command != "cp":
+        raise ValueError("Invalid command format. "
+                         "Usage: cp <source_file> <destination_file>")
+
+    if source_file == destination_file:
+        if not os.path.exists(source_file):
+            raise ValueError(f"Source file '{source_file}' does not exist")
```
YuriiKindrat, line 13: What if source_file and destination file will be the same but it will exist ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #323, opened 2023-02-19
Solution

VladyslavCherkashyn:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if "cp" not in command:
```
oigod, line 2: u can combine if`s

oigod (changes requested):
As i wrote in comments

app/main.py
VladyslavCherkashyn: could u clarify pls

oigod (approved):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, copied, file_to_copy = command.split(" ")
```
YuriiKindrat, line 2: no need to use space inside split method, it uses space by default, also use underscore only u won't use this variable later

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if copied == file_to_copy or _ != "cp":
+        return
```
YuriiKindrat, line 4: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,7 @@
...
+    with open(copied, "r") as file_in, open(file_to_copy, "w") as file_out:
+        for line in file_in:
+            file_out.write(line)
```
YuriiKindrat, line 7: use file.read() method instead

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, copied, file_to_copy = command.split()
+    if copied != file_to_copy or _ == "cp":
```
YuriiKindrat, line 3: use underscore for varuable only if u won't use this variable later, u are using it in if condition

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #325, opened 2023-02-20
Solution

ghost:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split(" ")
+    if files[0] != "cp" or files[1] == files[2]:
```
AlessaQ, line 3: Here you always have three words that make sense to unpack them into their own variables like `a, b, c = 1, 2, 3`

AlessaQ (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    task, first_file, second_file = command.split(" ")
```
Serhii-Leonenko, line 2: split() method use `" "` as a default value, so you can remove it

Serhii-Leonenko, line 2: I think `command` name will be better than `task` )

Serhii-Leonenko (changes requested):

wQuelS (approved):
GJ

PR #326, opened 2023-02-20
Solution

destr34:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split()
+    if open(files[1]) != open(files[2]) and files[0] == "cp":
```
ramsty, line 3: You're compering two items from list, no need to add "open"

app/main.py
```diff
@@ -1 +1,7 @@
...
+        with open(files[1], "r") as file_1, open(files[2], "w") as file_2:
+            df = file_1.readline()
+            for line in df:
+                file_2.write(line)
```
ramsty, line 7: no need to go line be line, you can write the whole file at once

ramsty (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split()
```
Serhii-Leonenko, line 2: You can unpack the list which we will get as the result of split() method, like this

`variable_a, variable_b, variable_c = "a b c".split()`

Serhii-Leonenko (changes requested):

wQuelS (approved):
GJ

PR #327, opened 2023-02-20
Solution

ada-krav:


destr34 (approved):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(file_names: str) -> None:
+    command, original, copy = file_names.split()
+    if command != "cp":
+        exit(f"Command '{command}' was not found")
+    if original == copy:
+        exit(0)
```
YuriiKindrat, line 6: no need to use exit here, u can just make a backward condition for code below

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(file_names: str) -> None:
+    command, original, copy = file_names.split()
+    if original == copy:
+        exit(0)
+    if command == "cp":
```
Serhii-Leonenko, line 5: the same, no need exit here, make one condition form this both with `and` operator

app/main.py
```diff
@@ -1 +1,8 @@
...
+        with open(original, "r") as original, open(copy, "w") as copy:
+            lines = original.readlines()
+            copy.writelines(lines)
```
Serhii-Leonenko, line 8: you can try to remove the redundant variable and make the same without it in one line

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Good job!

PR #328, opened 2023-02-20
Solution

oigod:


oksanaaam (approved):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, source_file, result_file = command.split()
+
+    if source_file == result_file or command != "cp":
+        return
+
+    with (open(source_file, "r") as source,
+          open(result_file, "w") as result):
+        result.write(source.read())
```
Serhii-Leonenko, line 9: You can do the same without return, just move the context manager inside the condition

Serhii-Leonenko (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #330, opened 2023-02-20
Solution

MikhailLyvak:


Anatolii-Poznyak (approved):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, old_file, new_file = command.split(" ")
```
YuriiKindrat, line 2: use full names for naming, command - not cmd

YuriiKindrat: also no need to pass space as arg in split method, it uses space by default

YuriiKindrat (changes requested):

Serhii-Leonenko (approved):
Well done!

PR #338, opened 2023-02-20
Solution

RomanRega:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(comand: str) -> None:
+    names = comand.split(" ")
+    if names[1] == names[2]:
+        print("Does nothing")
+    if names[1] != names[2]:
+        with open(names[1], "r") as file_in, open(names[2], "w") as file_out:
+            for i in file_in:
+                file_out.write(i)
```
ProdyRodion, line 8: U don't need for here

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(comand: str) -> None:
+    names = comand.split(" ")
+    if names[1] == names[2]:
```
ProdyRodion, line 3: u may do 1, 2, 3 = comand.split(" ") unstead using names[1] and ect.

ProdyRodion (changes requested):

ykopatko (approved):

app/main.py
Serhii-Leonenko: agree, you can unpack values into variables, and than use them it your code, but don't use `" "` inside of split(), because it is a default value

app/main.py
```diff
@@ -1 +1,8 @@
...
+        print("Does nothing")
```
Serhii-Leonenko, line 4: it's redundant condition and print 

image

app/main.py
Serhii-Leonenko: agree, do it without for loop

Serhii-Leonenko (changes requested):
Read the checklist 

image

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(comand: str) -> None:
+    names = comand.split()
```
Serhii-Leonenko, line 2: You can unpack the list which you will get from split() method like this:

`variable_a, variable_b, variable_c = "a b c".split()`

Serhii-Leonenko (changes requested):

PR #340, opened 2023-02-20
'Solution'

Bohdan-Kovalenko1992:


NickSavchyn (approved):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
```
Serhii-Leonenko, line 2: It's better to unpack the list which you get with split() like this
`variable_a, variable_b, variable_c = "abc".split()`

Serhii-Leonenko: and then use this variable in your code

Serhii-Leonenko (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
...
+    with (
+        open(command[1], "r") as file_to_copy,
+        open(command[2], "w") as new_file
+    ):
+        if file_to_copy.name != new_file.name and command[0] == "cp":
+            new_file.write(file_to_copy.read())
+        else:
+            print("Error: same file name.")
```
Serhii-Leonenko, line 10: make this check before you open the context manager, and remove redundant print with else condition

wQuelS (approved):
Nice

PR #341, opened 2023-02-20
Solution

tytomyr:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(string: str) -> None:
+    string = string.split(" ")
```
dogmoony, line 2: you don`t need to put " " inside function split, it will do that automaticly.

app/main.py
```diff
@@ -1 +1,11 @@
...
+    command = string[0]
+    file_name = string[1]
+    new_file_name = string[2]
+    if command == "cp" and file_name != new_file_name:
+        with open(f"{file_name}", "r") as file:
+            content = file.read()
+
+        with open(f"{new_file_name}", "w") as new_file:
+            new_file.write(content)
```
dogmoony, line 11: Do it by using one construction "with" as it was written in guide.

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(string: str) -> None:
+    string = string.split(" ")
+    command = string[0]
+    file_name = string[1]
+    new_file_name = string[2]
```
dogmoony, line 5: You can give values for these variables in one line, by using construction "a, b = "a b".split()"

dogmoony (changes requested):
work a little bit harder ;)

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(string: str) -> None:
+    command, file_name, new_file_name = string.split()
+    if command == "cp" and file_name != new_file_name:
+        with (open(f"{file_name}", "r") as file,
+             open(f"{new_file_name}", "w") as new_file):
```
dogmoony, line 5: you don`t need to use f-string, because file_name is already string
next line the same

dogmoony (changes requested):

dogmoony (approved):
Прекрасна робота!

YuriiKindrat (approved):

PR #346, opened 2023-02-20
Solution

Zenionoi:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    name_of_file = command.split()
+    if name_of_file[1] == name_of_file[2] or name_of_file[1][-4:] != ".txt":
```
Yulia-Mala, line 3: there is no condition in README to copy only txt files

app/main.py
```diff
@@ -1 +1,8 @@
...
+        return
+
+    from shutil import copyfile
```
Yulia-Mala, line 6: I see the first time import in the middle of the file, but maybe it's ok, I am not sure)

app/main.py
Yulia-Mala: I should reject because of it, pls fix it=)

Yulia-Mala (changes requested):

Yulia-Mala (approved):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    name_of_file = command.split()
+    if name_of_file[0] != "cp" or name_of_file[1] == name_of_file[2]:
+        return
+
+    from shutil import copyfile
```
wQuelS, line 6: Is your idea to do import only `if` statement is False?:) Cuz I don't get in what other cases could this import be here.
It certainly makes your code more unstructured. But  do you have any benefits?

wQuelS (changes requested):

app/main.py
Zenionoi: It was intended that way, but in reality there will be fewer cases where the function will stop on `if statement`, so bad  structure does not override by single absences of import

wQuelS (approved):
Good

PR #349, opened 2023-02-21
Solution

Esxoyne:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    args = command.split()
+    if len(args) != 3 or args[0] != "cp" or args[1] == args[2]:
```
Grayder0152, line 3: More correct create valiable instead of `args[0]`....

Grayder0152 (approved):

PR #351, opened 2023-02-22
solution

T-Vlad-96:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command_and_file_names: str) -> None:
+    source_file = command_and_file_names.split()[1]
+    new_file = command_and_file_names.split()[2]
```
olena-hul, line 3: Use variable to store command_and_file_names.split() result. Validate input command, whether it is cp or not and check command lenght (should be exacty 3 arguments)

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command_and_file_names: str) -> None:
+    command_and_file_names = command_and_file_names.split()
+    if "cp" in command_and_file_names and len(command_and_file_names) == 3:
+        source_file = command_and_file_names[1]
+        new_file = command_and_file_names[2]
```
ArsenPidhoretskyi, line 5: use correct name: destination_file_path

ArsenPidhoretskyi (approved):
Good job!

PR #352, opened 2023-02-22
solved

SDobzhynetskyi:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[1] == command[2]:
```
olena-hul, line 3: Use unpacking instead of indexes. Validate command lenght and command value (should be exactly "cp")

app/main.py
```diff
@@ -1 +1,7 @@
...
+        return
+    with open(command[1], "r") as file_in, open(command[2], "w") as file_out:
+        file1 = file_in.read()
```
olena-hul, line 6: Use more descriptive name instead of file1

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, file_name, file_name_copy = command.split()
```
ArsenPidhoretskyi, line 2: if len(command.split()) != 3 your code just broke without any info

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if command_name == "cp" and file_name != file_name_copy:
+        with open(file_name, "r") as file_in, \
+                open(file_name_copy, "w") as file_out:
```
```suggestion
        with (
            open(file_name, "r") as file_in, 
            open(file_name_copy, "w") as file_out
        ):
```

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

PR #353, opened 2023-02-22
Solution

lotrik84:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    original_file = command.split()[1]
+    new_file = command.split()[2]
```
Vasyl-Smutok, line 3: you can do it like this
```suggestion
    command_name, base_filename, new_filename = command.split()
```

app/main.py
```diff
@@ -1 +1,8 @@
...
+
+    if original_file != new_file:
```
Vasyl-Smutok, line 5: also add check for this case  
image

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, original_file, new_file = command.split()
```
ArsenPidhoretskyi, line 2: validate that command has exactly 3 parameters

app/main.py
```diff
@@ -1 +1,7 @@
...
+
+    if command_name == "cp" and original_file != new_file:
+        with (open(original_file, "r") as file_in,
+              open(new_file, "w") as file_out):
```
```suggestion
        with (
            open(original_file, "r") as file_in,
            open(new_file, "w") as file_out
        ):
```

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #354, opened 2023-02-23
Solution

tarasavage:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, file, file_copy = command.split(" ")
```
Grayder0152, line 2: You should check if `command` length == 3, before it

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split(" ")
+    if len(parts) == 3:
+        command_name, file, file_copy = parts
```
ArsenPidhoretskyi, line 4: file, file_copy not clear name
first of all it is source_file_path, destination_file_path

app/main.py
```diff
@@ -1 +1,8 @@
...
+        if file != file_copy and command_name == "cd":
+            with (open(file, "r") as file_in,
+                  open(file_copy, "w") as file_out):
```
```suggestion
            with (
               open(file, "r") as file_in,
               open(file_copy, "w") as file_out
            ):
```

ArsenPidhoretskyi: rename file_in, file_out to source_file, destination_file

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #355, opened 2023-02-24
Solution

Roman28101:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split = command.split(" ")
```
ArsenPidhoretskyi, line 2: split is not clear name
also you must validate before that command.split() has exactly 3 elements

app/main.py
```diff
@@ -1 +1,14 @@
...
+    if split[1] != split[2] and split[0] == "cp":
+        try:
+            with (open(split[1], "r") as file_in,
+                  open(split[2], "w") as file_out
+                  ):
+                for line in file_in:
+                    file_out.write(line)
```
ArsenPidhoretskyi, line 9: Can you try:

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split = command.split(" ")
+    if split[1] != split[2] and split[0] == "cp":
+        try:
+            with (open(split[1], "r") as file_in,
+                  open(split[2], "w") as file_out
+                  ):
```
```suggestion
            with (
                open(split[1], "r") as file_in,
                open(split[2], "w") as file_out
            ):
```

ArsenPidhoretskyi (changes requested):

Roman28101:
> Good Job 👍
> 
> But `sep_command` also bad variable name what does `sep` mean?

Actually its "separate". I will try another naming

Vasyl-Smutok (approved):
Well Done 👍

PR #356, opened 2023-02-24
Solution

WalDragnir:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
```
ArsenPidhoretskyi, line 2: validate that command have exactly 3 parameters

ArsenPidhoretskyi: Also you need to validate if it cp command
Read checklist

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #357, opened 2023-02-25
Add solution

ZaikaDaria:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    line = command.split()
```
Grayder0152, line 2: Use unpacking for this case, but before chach that `len(command) == 3`

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if line[0] != "cp" or line[1] == line[2]:
+        pass
+    else:
+        with open(line[1], "r") as file_in, open(line[2], "w") as file_out:
+            file_out.write(file_in.read())
```
Grayder0152, line 7:
```suggestion
    if line[0] == "cp" and line[1] != line[2]:
        with open(line[1], "r") as file_in, open(line[2], "w") as file_out:
            file_out.write(file_in.read())
```

Grayder0152 (changes requested):

app/main.py
ArsenPidhoretskyi: Strange checks:
If len(line) != 3 then go inside - wrong!
if line[0] != "cp" then do copy - wrong!
if line[1] == line[2] then  do copy - wrong!


All checks are wrong! Must be all 3 true len == 3, the command is cp, files names are different

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    line = command.split()
```
katevasilenko, line 2: Also you can unpack command to variables right here and than use them in code. It makes your code more readable 
```
    variable, variable, variable = command.split()
```

katevasilenko (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, filename, new_filename = command.split()
+    if len(command.split()) == 3 or cmd == "cp" or filename != new_filename:
```
ArsenPidhoretskyi, line 3: You need to validate len(command.split()) == 3 before unpacking(line 2)

app/main.py
```diff
@@ -1 +1,6 @@
...
+        with open(filename, "r") as file_in, \
+             open(new_filename, "w") as file_out:
```
```suggestion
        with (
            open(filename, "r") as file_in,
            open(new_filename, "w") as file_out
        ):
```

app/main.py
ArsenPidhoretskyi: Your code will return unknown error now if command will be "cp file.txt"

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #358, opened 2023-02-26
Solution

Andrew13X:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.split()[1] != command.split()[2]:
```
olena-hul, line 2: Move command.split() to a variable
Validate variable length
Validate that first argument is exactly "cp" command
And use unpacking instead of index access

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file, new_file = command.split()
```
ArsenPidhoretskyi, line 2: Before unpacking you need to validate command have exactly 3 parameters
Also name cp, file, new_file is not clear names
cp -> input_command
file -> source_file_path
new_file -> destination_file_path

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if cp == "cp" and file != new_file:
+        with open(file, "r") as original_file, \
+                open(new_file, "w") as file_copy:
```
```suggestion
        with (
            open(file, "r") as original_file,
            open(new_file, "w") as file_copy
        ):
```

ArsenPidhoretskyi (changes requested):

app/main.py
Andrew13X: got it

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3:
+        input_command, source_file_path, destination_file_path \
```
Grayder0152, line 3: Use `()` instead of `\`

app/main.py
```diff
@@ -1 +1,8 @@
...
+            = command.split()
+        if input_command == "cp" and source_file_path != destination_file_path:
+            with open(source_file_path, "r") as original_file, \
+                    open(destination_file_path, "w") as file_copy:
```
Grayder0152, line 7: the same as above

Grayder0152 (changes requested):

app/main.py
Andrew13X: done

app/main.py
```diff
@@ -1 +1,12 @@
...
+        (
+            input_command,
+            source_file_path,
+            destination_file_path
+        ) = command.split()
+        if input_command == "cp" and source_file_path != destination_file_path:
+            with open(source_file_path, "r") as original_file, open(
+                    destination_file_path, "w"
+            ) as file_copy:
+                file_copy.write(original_file.read())
```
Vasyl-Smutok, line 12:
```suggestion
            with (
                open(source_file_path, "r") as original_file, 
                open(destination_file_path, "w") as file_copy
            ):
                file_copy.write(original_file.read())
```

Vasyl-Smutok (approved):
Well Done 👍🏻

PR #359, opened 2023-02-27
Solution

yhwh6:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, current_file, copied_file = command.split()
```
Grayder0152, line 2: You must check that `len(command) == 3` before it

Grayder0152:
```suggestion
    cp, current_file_name, copied_file_name = command.split()
```

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if cp != "cp":
+        return
+    with (open(f"{current_file}.txt", "r") as file_to_copy,
+          open(f"{copied_file}.txt", "w") as file_result):
+        file_result.write(file_to_copy.read())
```
Grayder0152, line 7:
```suggestion
    if cp == "cp" and currect_file_name != copied_file_name:
        with (
            open(f"{current_file}.txt", "r") as file_to_copy,
            open(f"{copied_file}.txt", "w") as file_result
        ):
             file_result.write(file_to_copy.read())
```

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command) == 3:
```
ArsenPidhoretskyi, line 2: Not correct check because "hello" will return true in this case
Command must contain THREE ARGUMENTS

app/main.py
```diff
@@ -1 +1,10 @@
...
+        cp, current_file_name, copied_file_name = command.split()
```
ArsenPidhoretskyi, line 3: cp -> input_command

ArsenPidhoretskyi (changes requested):

PR #361, opened 2023-03-05
'Solution'

arsenmakovei:


ArsenPidhoretskyi (changes requested):
1. You need to validate that command have exactly 3 parameters
2. https://github.com/mate-academy/py-copy-file#note-check-your-code-using-this-checklist-before-pushing-your-solution not read as i see

ArsenPidhoretskyi (approved):
Good job!

PR #364, opened 2023-03-10
'Solution'

StinGER-ua:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file, new_file = command.split()
```
ArsenPidhoretskyi, line 2: validate that command.split() will have exactly 3 params

ArsenPidhoretskyi: file, new_file is not clear name, because it is not a files, it path

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if command == "cp" and file != new_file:
+        with (open(file, "r") as file_in,
+              open(new_file, "w") as file_out):
```
```suggestion
        with (
            open(file, "r") as file_in,
            open(new_file, "w") as file_out
        ):
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_name, new_file_name = command.split()
```
Vasyl-Smutok, line 2: this line can cause an error, I think it should be swapped with your check

Vasyl-Smutok (changes requested):

app/main.py
StinGER-ua: I don't know what to do. Please advise.

StinGER-ua: > this line can cause an error, I think it should be swapped with your check

I don't know what to do. Please advise.

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3:
+        command, file_name, new_file_name = command.split()
```
ArsenPidhoretskyi, line 3: Otherwise do return.

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #365, opened 2023-03-10
Solution

kostya-kononenko:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_of_command = list(command.split())
+    if list_of_command[1] == list_of_command[2]:
```
vasylhnatiuk, line 3: If len(command.split()) less then 3, your code will broke

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_of_command = list(command.split())
```
vasylhnatiuk, line 2: would be better to use `command, copied_file, file_to_copy = command.split()` for split

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if list_of_command[1] == list_of_command[2]:
+        return None
+    else:
```
vasylhnatiuk, line 5: else is useless here

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, copied_file, file_to_copy = command.split()
```
Vasyl-Smutok, line 2: validate that command.split() will have exactly 3 params

Vasyl-Smutok (changes requested):

app/main.py
kostya-kononenko: Validate in code? I understand that according to the conditions of the task there will be only 3 parameters, or I'm mistaken? "Function take only one argument command, that is string with **command cp**, **file name to copy** and **new file name**, separated by spaces."

vasylhnatiuk: yes, but if command is wrong or has only one argument, like command = "cp txt.txt" or = "file.txt" function will broke. You can validate this : if len(command.split()) == 3

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: change order

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,7 +1,7 @@
 def copy_file(command: str) -> None:
-    command, copied_file, file_to_copy = command.split()
     if len(command.split()) == 3:
-        if copied_file != file_to_copy:
-            with open(copied_file, "r") as file_in, \
-                    open(file_to_copy, "w") as file_out:
-                file_out.write(file_in.read() + "\n")
+        command, copied_file, file_to_copy = command.split()
+        if command == "cp" and copied_file != file_to_copy:
+                with open(copied_file, "r") as file_in, \
```
vasylhnatiuk: `app/main.py:5:17: E117 over-indented ` please fix flake8 error, test failed

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):
Good job!

PR #370, opened 2023-03-20
'Solution'

iishchenko:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+from shutil import copy
+
+
+def copy_file(command: str) -> None:
+    _, filename1, filename2 = command.split()
```
Grayder0152, line 5: Pls, check that command length == 3, and command is `cp`, and file_1 != file_2

Grayder0152 (changes requested):

PR #374, opened 2023-03-27
Solution

MaksNochvai:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parameters = command.split(" ")
```
olena-hul, line 2: No need to pass `" "` inside split, because it is a default separator

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    if len(parameters) == 3 \
+            and parameters[0] == "cp" \
+            and parameters[1] != parameters[2]:
```
olena-hul, line 6: Use unpacking instead of using indexes to access command parts. And use `()` to split lines

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parameters = command.split()
+
+    if (len(parameters) == 3
+            and parameters[0] == "cp"
+            and parameters[1] != parameters[2]):
+        new_file = parameters[2]
+        old_file = parameters[1]
```
Bohdan-Salamakha, line 8:
```
new_file = parameters[2]
old_file = parameters[1]
```

add this code before "if" condition to not duplicate calls parameters[1] and parameters[2]

Bohdan-Salamakha (changes requested):
v1

app/main.py
ArsenPidhoretskyi: @MaksNochvai not fixed

ArsenPidhoretskyi: Also before getting by index element you must validate that command have 3 parameters

ArsenPidhoretskyi (changes requested):
Check your code with command:
"cp"
If it raise an error?

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parameters = command.split()
+    if len(command.split()) == 3:
+        parameters = command.split()
```
Bohdan-Salamakha, line 4: what is the sense of it?
```suggestion
```

Bohdan-Salamakha (changes requested):
v2

app/main.py
olena-hul: The check for the length should be present. Instead you shouldn't duplicate command.split() call. Save the result of this call to a variable and then use unpacking

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parameter, old_file, new_file = command.split()
```
olena-hul, line 2: Please return the check for the length

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3:
+        parameter, old_file, new_file = command.split()
```
EdAlekseiev, line 3: Please, don't split `command` twice. Create a variable for storing the result `command.split()` and use it on demand.

Bohdan-Salamakha (changes requested):
fix an issue
image

app/main.py
```diff
@@ -1 +1,13 @@
...
+def copy_file(command: str) -> None:
+    split_command = command.split()
+    if len(split_command) == 3:
+        parameter, old_file, new_file = split_command
+
+        if parameter != "cp":
+            raise ValueError("Parameter must be 'cp'")
+
+        if old_file == new_file:
+            raise ValueError("old_file and new_file must have different names")
```
Bohdan-Salamakha, line 10: It must do nothing
U don't need to raise any errors
https://github.com/mate-academy/py-copy-file#:~:text=It%20must%20do%20nothing%20in%20case%20the%20user%20is%20trying%20to%20copy%20file%20to%20file%20with%20the%20same%20name.
image

Bohdan-Salamakha (changes requested):
v3

Bohdan-Salamakha (changes requested):
program doesn't copy file 
image
https://github.com/mate-academy/py-copy-file#:~:text=copy_file(%22cp%20file.txt%20new_file.txt%22)

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
+    if len(split_command) == 3:
+        parameter, old_file, new_file = split_command
+
+        if (parameter != "cp"
+                and old_file != new_file):
```
```suggestion
        if (parameter != "cp" and old_file != new_file):
```
Format as follow

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
+    if len(split_command) == 3:
+        parameter, old_file, new_file = split_command
```
ArsenPidhoretskyi, line 4: It is file_path not file
old_file, new_file

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+        if (parameter != "cp"
+                and old_file != new_file):
+            with (open(old_file, "r") as file_in,
+                  open(new_file, "w") as file_out):
```
```suggestion
            with (
               open(old_file, "r") as source,
               open(new_file, "w") as destination
           ):
```

ArsenPidhoretskyi (approved):
Good job, but please use corrrect naming

PR #375, opened 2023-03-29
'Solution'

dominyevhen:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split()
+    if commands[1] != commands[2]:
+        with open(commands[1], "r") as file_in,\
+                open(commands[2], "w") as file_out:
```
Bohdan-Salamakha, line 5: відформатуй по ти цього
```
with (open(commands[1], "r") as file_in,
      open(commands[2], "w") as file_out):
```

Bohdan-Salamakha, line 5: what are commands[1], commands[2]?
create separated variables for them and call on demand

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    current_file = command.split()[1]
+    new_file = command.split()[2]
```
Bohdan-Salamakha, line 3: don't split your command twice
create a separated variable for command.split() and use it on demand

Bohdan-Salamakha (changes requested):
v2

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, current_file, new_file = command.split()
+    if current_file != new_file:
+        with (open(current_file, "r") as file_in,
+              open(new_file, "w") as file_out):
+            file_out.write(file_in.read())
```
EdAlekseiev, line 6: Please, add checking that command is correct:
1. Starts from cp
2. Contains all arguments

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, current_file, new_file = command.split()
```
olena-hul, line 2: Please check that command length is equal exactly to 3 arguments

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if (current_file != new_file
+            and command == "cp" and (current_file, new_file)):
```
olena-hul, line 4: ```and (current_file, new_file)):```

Can you explain why did you use this construction and how does this work?

olena-hul (changes requested):

app/main.py
dominyevhen: current_file is True, new_file is True I mean

Bohdan-Salamakha (approved):

PR #376, opened 2023-03-30
Solution

AlexandrZhezheria:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(cp: str) -> None:
```
ArsenPidhoretskyi, line 1: cp? Not clear name

app/main.py
```diff
@@ -1 +1,9 @@
...
+    tokens = cp.split()
```
ArsenPidhoretskyi, line 2: tokens? жетони?

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if len(tokens) != 3:
+        raise ValueError("Invalid command")
+    source_file, dest_file = tokens[1], tokens[2]
+    if source_file == dest_file:
+        return
```
ArsenPidhoretskyi, line 7: If command will equal to "cc" we also do copy?
https://github.com/mate-academy/py-copy-file/blob/master/checklist.md

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Good job!

PR #377, opened 2023-04-01
'Solution'

HalynaPetrova:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    new_file = command.split(" ")
+    if new_file[1] != new_file[2]:
+        with open(new_file[1], "r") as file_in, \
+                open(new_file[2], "w") as file_out:
```
Grayder0152, line 5: Use `()` instead of `\`

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    new_file = command.split(" ")
+    if new_file[1] != new_file[2]:
```
Grayder0152, line 3: before it pls check that len of `command.split()` == 3 and command name == `cp`

Grayder0152 (changes requested):

Grayder0152 (approved):

PR #380, opened 2023-04-02
Solution

vlamuha:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_of_command = command.split(" ")
+    if list_of_command[0] == "cp" and list_of_command[1] != list_of_command[2]:
```
Grayder0152, line 3: before it check that `command len == 3`

app/main.py
```diff
@@ -1 +1,6 @@
...
+        with open(list_of_command[1], "r") as f_in,\
+             open(list_of_command[2], "w") as f_out:
```
Grayder0152, line 5: use `()` instead of `\`

Vasyl-Smutok (approved):
Good job 👍 

but,
`f_in`, `f_out` bad variable names

PR #381, opened 2023-04-03
Solution

V-Shkrobatskyi:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    sp_com = command.split()
```
Vasyl-Smutok, line 2: Before unpacking you need to validate command have exactly 3 parameters

Vasyl-Smutok (changes requested):

Vasyl-Smutok:
https://github.com/mate-academy/py-copy-file/pull/381#discussion_r1156338162

not fixed

Vasyl-Smutok (changes requested):
https://github.com/mate-academy/py-copy-file/pull/381#discussion_r1156338162

not fixed

V-Shkrobatskyi:
```
    words = command.split(" ")
    if len(words) != 3 or words[0] != "cp":
        return None
```
something like this?

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    words = command.split(" ")
+    if len(words) != 3 or words[0] != "cp":
+        return None
+    if words[1] != words[2]:
```
Grayder0152, line 5:
```suggestion
    if len(words) == 3 and words[0] == "cp" and words[1] != words[2]:
  
```

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    words = command.split(" ")
```
Grayder0152, line 2: pls unpack result of split to 3 variables

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
...
+    command_name = words[0]
```
Vasyl-Smutok, line 3: if len(command.split()) == 3:
        command_name, name_to_copy, new_file_name = command.split()

Vasyl-Smutok (changes requested):

app/main.py
V-Shkrobatskyi: ah, so this is the way

V-Shkrobatskyi: I was trying to avoid double if

Vasyl-Smutok (approved):
👍

PR #382, opened 2023-04-03
Solution

nickkozlov90:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_names = command.split()
+    if files_names[1] == files_names[2]:
+        return
+    with open(f"{files_names[1]}", "r") as file_in, \
+            open(f"{files_names[2]}", "w") as file_out:
```
Bohdan-Salamakha, line 6: what are files_names[1], files_names[2]?
create separated variables for their values

Bohdan-Salamakha, line 6: it is better to wrap lines using parentheses
```
with (open(f"{files_names[1]}", "r") as file_in,
      open(f"{files_names[2]}", "w") as file_out):

```

Bohdan-Salamakha (changes requested):
v1

PR #386, opened 2023-04-10
'Solution#copyFile'

Jhaivir:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    part_of_command = command.split(" ")
+    if part_of_command[0] == "cp":
```
ArsenPidhoretskyi, line 3: First of all, validate that command have 3 parameters

app/main.py
```diff
@@ -1 +1,11 @@
...
+        if part_of_command[1] == part_of_command[2]:
+            return
```
ArsenPidhoretskyi, line 5: You can add this check to upper if

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #387, opened 2023-04-10
'Solution'

OleksandrYanchuk:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split()
+    old_file = command_split[1]
+    new_file = command_split[2]
```
ArsenPidhoretskyi, line 4: Move it after validation length of command

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if len(command_split) != 3 or command_split[0] != "cp":
+        return
+    if old_file == new_file:
+        return
```
ArsenPidhoretskyi, line 8:
```suggestion
    if len(command_split) != 3:
        return
    command, source_path, destination_path = command_split
    if command != "cp" or source_path == destination_path:
        return
```

app/main.py
```diff
@@ -1 +1,10 @@
...
+    with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+        file_out.write(file_in.read())
```
ArsenPidhoretskyi, line 10: rename file_in to source_file
rename file_out to destination_file

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if len(command_split) != 3:
+        return
+    command, source_path, destination_path = command_split
+    if command != "cp" or source_path == destination_path:
+        return
+    with (
+        open(source_path, "r") as source,
+        open(destination_path, "w") as destination
+    ):
+        source.write(destination.read())
```
ArsenPidhoretskyi, line 12: Wrong order you must write inside destination not inside source

ArsenPidhoretskyi (changes requested):

Bohdan-Salamakha (approved):

PR #393, opened 2023-04-18
Solution

Maksym-Filipchuk:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_in = command.split(" ")[1]
```
ArsenPidhoretskyi, line 2: 1. First of all you must check that command have exactly 3 parameters
2. Also space is the default argument in .split
3. You should read checklist:
https://github.com/Maksym-Filipchuk/py-copy-file/blob/develop/checklist.md#code-efficiency

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_type = command.split()[0]
```
ArsenPidhoretskyi, line 2: 1. Splitting 3 times is not a good idea
2. You can do:
```
command = command.split()
# do validation that command have 3 arguments
command_type, source_path, copy_path = command
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):
Well done!

PR #394, opened 2023-04-19
Solution

Fawnygod:


app/main.py
```diff
@@ -1 +1,7 @@
 # write your code here
+def copy_file(command: str) -> None:
+    old_file = command.split()[1]
+    new_file = command.split()[2]
```
ArsenPidhoretskyi, line 4: First of all validate that command have 3 parameters

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
 # write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3:
+        old_file = command.split()[1]
+        new_file = command.split()[2]
```
```suggestion
        command, old_file, new_file = command.split()
```
No need to split three times

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #395, opened 2023-04-20
Develop

kucherenkodmytro:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    new_command = command.split()
```
vasylhnatiuk, line 2: add check if len(command.split()) != 3 or command.split()[0] != "cp":

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if (
+            new_command[1] != new_command[2]
+            and len(command.split()) == 3
```
ArsenPidhoretskyi, line 5: Checking for length needs to be first because if it has not 3 arguments it will break when you try to do new_command[2]

ArsenPidhoretskyi (changes requested):

PR #396, opened 2023-04-24
Solution of copy file

svyshcho:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+
+
+def copy_file(command: str) -> None:
+    names = command.split()
+    if names[1] == names[2] or\
+            len(command.split()) != 3 or\
+            command.split()[0] != "cp":
```
olena-hul, line 7: Check command length first
Then use unpacking like `command, source_file, target_file = names`
Then check command name and files

app/main.py
```diff
@@ -1 +1,18 @@
...
+        return
+    with open(names[1], "r") as f_r, open(names[2], "w") as f_w:
```
olena-hul, line 9: `f_r` and `f_w` are not descriptive variable names

app/main.py
```diff
@@ -1 +1,18 @@
...
+        arr_data = f_r.readlines()
+        arr_data_checked = []
+        for i in arr_data:
+            arr_data_checked.append(i.strip("\n"))
+        for line in arr_data_checked:
+            f_w.write(line + "\n")
```
olena-hul, line 15: Why do you remove `\n` symbol and then add it again?

You can just read the content of the one file and write it to another file without any modifications

app/main.py
```diff
@@ -1 +1,18 @@
...
+
+
+copy_file("cp test1.txt file8.txt")
```
olena-hul, line 18: Remove this line

app/main.py
svyshcho: Done

app/main.py
svyshcho: Done

app/main.py
svyshcho: Done

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+
+
+def copy_file(command: str) -> None:
+    names = command.split()
+    if names[1] == names[2] or\
+            len(command.split()) != 3 or\
+            command.split()[0] != "cp":
```
EdAlekseiev, line 7: You must change the order of checks:
1. Check that command `names` contains only 3 elements
2. Check that first element of `names` is "cp"
3. Check that the old name of file doesn't equal to the new name
Also, don't use command.split() a few times, please.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #399, opened 2023-04-30
done

chipapapo:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    old_file, new_file = command.split(" ")[1:]
```
Vasyl-Smutok, line 2: add a check that the command was entered correctly

Vasyl-Smutok (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, old_file, new_file = command.split(" ")
```
vasylhnatiuk, line 2: If `command does not have 3 arguments, the code will be broken. Add a check for it before

app/main.py
```diff
@@ -1 +1,5 @@
...
+    if cp == "cp" and old_file != new_file:
+        with open(old_file, "r") as old, open(new_file, "w") as new:
```
vasylhnatiuk, line 4: `new `and `old ` are bad names

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #400, opened 2023-05-01
Solution

MaksymUK:


app/main.py
```diff
@@ -1 +1,13 @@
 # write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split(" ")
```
olena-hul, line 3:
```suggestion
    split_command = command.split()
```

Space is a default argument in `split`

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if len(split_command) < 3 or split_command[1] == split_command[2]:
```
olena-hul, line 4: You can use unpacking instead of accessing list items by indexes. For example: a, b = [1,2]

olena-hul: Add a check for command name, ensure that it equals exactly to "cp"

app/main.py
```diff
@@ -1 +1,13 @@
...
+        return
```
olena-hul, line 5: Return an error message or print it to indicate that something has gone wrong

app/main.py
```diff
@@ -1 +1,13 @@
...
+    else:
```
olena-hul, line 6: `else` is not needed after `return` statement

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> str:
+    cp_command, file_in_name, file_out_name = command.split()
+
+    if (
+            len(command.split()) < 3
```
Bohdan-Salamakha, line 5: don't duplicate command "command.split()" twice
create a separated variable for it and call it on demand

app/main.py
```diff
@@ -1 +1,16 @@
...
+            or file_in_name == file_out_name
+            or cp_command != "cp"
+    ):
+        return "Something has gone wrong. Please check command entry"
+    with open(
+            file_in_name, "r"
+    ) as file_in, open(
+        file_out_name, "w"
+    ) as file_out:
```
Bohdan-Salamakha, line 14: it's hard to read
format it better

Bohdan-Salamakha (changes requested):
v1

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> str:
+    split_command = command.split()
+    cp_command, file_in_name, file_out_name = split_command
```
olena-hul, line 3: If the command length will be less than 3, the error will be risen

You should check the length first, and then unpack values
Use 2 `if` stamements

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+    if (
+            len(split_command) < 3
+            or file_in_name == file_out_name
+            or cp_command != "cp"
+    ):
+        return "Something has gone wrong. Please check command entry"
+
+    with open(file_in_name, "r") as file_in,\
+            open(file_out_name, "w") as file_out:
```
olena-hul, line 13: Don't use `\` to split lines, it is a bad practice

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
...
+    cp_command, file_in_name, file_out_name = split_command
+
+    if (
+            file_in_name == file_out_name
+            or cp_command != "cp"
+    ):
+        return "Something has gone wrong. Please check command entry"
+
+    with open(file_in_name, "r") as file_in:
+        with open(file_out_name, "w") as file_out:
```
olena-hul, line 15:
```suggestion
    with (
         open(file_in_name, "r") as file_in,
         open(file_out_name, "w") as file_out
    ):
```

olena-hul (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #401, opened 2023-05-03
solution

alim2709:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    new_list = command.split()
+    if new_list[1] == new_list[2]:
+        return
+    with (
+        open(new_list[1], "r") as file_need_to_copy,
+        open(new_list[2], "w") as copied_file
```
Bohdan-Salamakha, line 7: what are new_list[1], new_list[2]?
create separated variables for them and call them on demand

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    new_list = command.split()
+    file_needs_copy = new_list[1]
+    new_copy_of_file = new_list[2]
```
EdAlekseiev, line 4: Use unpacking instead of accessing list items by indexes

EdAlekseiev: Add check that command starts from "cp"

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #403, opened 2023-05-05
Solution

OlehOryshchuk:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    text = command.split(" ")
```
EdAlekseiev, line 2: `" "` is default value for `split`

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    text = command.split()
+    cp, from_file, to_file = text
+    if "cp" not in cp or len(text) != 3:
```
EdAlekseiev, line 4:
```suggestion
    if "cp" != cp or len(text) != 3:
```

EdAlekseiev: `not in` operator uses to determine if a given value isn’t part of a collection of values. In your case, you must compare these two values.

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #407, opened 2023-05-08
solution

neostyle88:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split()
```
ArsenPidhoretskyi, line 2: First of all: Checklist was not readed
https://github.com/neostyle88/py-copy-file/blob/develop/checklist.md

1. Files is not correct name.
2. Check that commands.split() have exactly 3 params
3. Than you can use unzipping:
cmd, sourse_path, destination_path = command.split()
4. Check that cmd is equal to cp
5. Before opening files you can check that names are not the same. Not need to do that inside context manager

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if len(command.split()) != 3:
+        return "Incorrect command is printed"
+    cmd, source_path, destination_path = command.split()
+    if cmd != "cp":
+        return "Command is not exist"
+    if source_path == destination_path:
+        return "The name of copied file is the same as original name"
+    with open(source_path, "r") as original, open(
+        destination_path, "w"
+    ) as copied:
```
olena-hul, line 11: original and copied are not very descriptive variable names. Use `source_file` and `target_file` for example

olena-hul:
Good job!

PR #409, opened 2023-05-08
Solution

VladyslavKaminskyi:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(comnand: str) -> None:
+    splited_command = comnand.split()
+    file_original = splited_command[1]
+    file_duplicate = splited_command[2]
```
Grayder0152, line 4: pls check that `splited_command` len == 3

app/main.py
```diff
@@ -1 +1,11 @@
...
+    if file_original == file_duplicate:
+        print("Can't copy file to file with the same name.")
+        return
+    with open(file_original, "r") as original,\
+            open(file_duplicate, "w") as duplicate:
```
Grayder0152, line 9: use () instead of \

Grayder0152 (changes requested):

PR #411, opened 2023-05-11
Solution

markus2812:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(cp: str) -> None:
+    without_cp = cp.split()
```
ArsenPidhoretskyi, line 2: Why without_cp named?

app/main.py
```diff
@@ -1 +1,10 @@
...
+    file_1 = without_cp[1]
```
ArsenPidhoretskyi, line 3: Check that command.split have 3 parameters

app/main.py
```diff
@@ -1 +1,10 @@
...
+    file_2 = without_cp[2]
+    if file_1 == file_2:
```
ArsenPidhoretskyi, line 5: Also you must check that command is really cp

app/main.py
```diff
@@ -1 +1,10 @@
...
+        return
+    with open(file_1, "r") as file1:
+        with open(file_2, "w") as file2:
```
ArsenPidhoretskyi, line 8: image
Do like this

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if "cp" not in command:  # перевірка на наявність команди cp в command
+        return
```
ArsenPidhoretskyi, line 3: It will not work properly:
for example: command == "hm copy.txt cpd.txt"
Will return true
And it will be wrong because command is hm

app/main.py
```diff
@@ -1 +1,12 @@
...
+    split_command = command.split()
+    print(split_command)  # перевірка на наявність 3 аргументів в split_command
+    file_1 = split_command[1]
+    file_2 = split_command[2]
```
ArsenPidhoretskyi, line 7: file_1, file_2 is not clear names choose better names

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if file_1 == file_2:
+        return
+    with open(file_1, "r") as file_in, open(file_2, "w") as file_out:
+        for line in file_in:
+            file_out.write(line)
```
ArsenPidhoretskyi, line 12: You can use file_in.write(file_out.read()) instead of for loop

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
+    if split_command[0] != "cp":
+        return
+    print(split_command)  # перевірка на наявність 3 аргументів в split_command
```
Grayder0152, line 5: It is not checking, use `if` to check that split_command len == 3

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
+    if len(split_command) != 3:
+        return
+    print(split_command)  # перевірка на наявність 3 аргументів в split_command
```
vasylhnatiuk, line 5: delete useless print

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
+    if len(split_command) != 3:
```
```suggestion
    if len(split_command) != 3 or split_command[0] != "cp":
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #417, opened 2023-05-17
py-copy-file

muriakula:


Bohdan-Salamakha (changes requested):
invalid input format
do it as u see in example
image

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_in = command.split(" ")[1]
+    file_out = command.split(" ")[2]
```
EdAlekseiev, line 3: Don't split `command` twice. Create a separate variable for storing result of `command.split()` and use it where you need.

EdAlekseiev, line 3: `" "` is default value `split`, you can use just `command.split()`

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if file_in != file_out:
```
EdAlekseiev, line 4: Also, check that command starts from `cp`.

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    elements = command.split()
```
olena-hul, line 2: Check command length, it should equal exactly to 3 arguments
Then use unpacking, for example:
a, b, c = [1,2,3]
This way you'll avoid accessing command items using indexes

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #423, opened 2023-05-19
COPY FILE

volodinsy:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    part = command.split()
+    first_file = part[1]
+    second_file = part[2]
```
Vasyl-Smutok, line 4: make sure it doesn't cause an error

Vasyl-Smutok (changes requested):
image

PR #428, opened 2023-05-22
'Solution'

Artistgatehub:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    copy_file_command = command_parts[0]
+    sour_file = command_parts[1]
+    des_file = command_parts[2]
```
EdAlekseiev, line 5: Use unpacking instead of accessing list items by indexes

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    if sour_file == des_file:
```
EdAlekseiev, line 7: Add check that command starts from "cp"

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    copy_file_command, sour_file, des_file = command.split()
```
YukihiroSM, line 2: What if the command is something like cp file.txt (only 2 parts)?
Add some check for elements amount.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
+        return
+    copy_file_command, sour_file, des_file = command.split()
```
Bohdan-Salamakha, line 4: don't split command twice
create a separate variable and use it on demand

app/main.py
```diff
@@ -1 +1,11 @@
...
+
+    if sour_file == des_file or copy_file_command != "cp":
+        return
+    else:
+        print(f"{copy_file_command} {sour_file} {des_file}")
```
Bohdan-Salamakha, line 9: remove test print

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
+        return
+    copy_file_command, sour_file, des_file = command.split()
+
+    if sour_file == des_file or copy_file_command != "cp":
+        return
+    else:
```
Bohdan-Salamakha, line 8: remove redundant "else"

Bohdan-Salamakha (changes requested):

PR #433, opened 2023-05-26
Developer

flatchers:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
```
vsmutok, line 2: Space by default 
```suggestion
    command = command.split()
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+    source_file = command[1]
+    destination_file = command[2]
```
vsmutok, line 4: check if this doesn't cause an error `IndexError: list index out of range`

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if source_file == destination_file:
+        return
+    with (open(source_file, "r") as file_in,
+            open(destination_file, "w") as file_out):
+        file_out.write(file_in.read())
+
+
+print(copy_file("cp file.txt new_file.txt"))
```
vsmutok, line 12: why print?

also add` if __name__ == "__main__":
`

vsmutok (changes requested):

PR #435, opened 2023-05-28
Dev

flatchers:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    source_file = command[1]
+    destination_file = command[2]
+    if len(command) != 2:
```
ArsenPidhoretskyi, line 5: image
First of all. Command is always 3 params
Also your code will not work if it comes with 2 params because you do [2]
Check that command is == "cp"

ArsenPidhoretskyi (changes requested):

PR #436, opened 2023-05-29
solution

Davison322:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    cp = command_list[0]
```
```
    cp = command_list[0]
    source_file = command_list[1]
    dest_file = command_list[2] 
```
it must be after len(command_list) == 3, because code will be broken if len(command_list) != 3

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if len(command_list) == 3:
+        cp = command_list[0]
+        src_file = command_list[1]
+        dst_file = command_list[2]
```
```suggestion
        cp, src_file, dst_file = command_list
```
You can unzip it

ArsenPidhoretskyi: Also rename, because for naw it i not good names:
cp -> cmd
src_file -> src_path
dst_file -> dst_path

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #440, opened 2023-05-30
Dvlp

flatchers:


flatchers (commented):
I think that i must to do main command after we wrote "cp" because cp in this task not only as text, it main command which provide coping and creating new file.

flatchers (commented):
/

flatchers (commented):
final changes

PR #441, opened 2023-05-30
implemented copy file

S-SMTN:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, file, copy = command.split(" ")
```
YukihiroSM, line 2: what if command is not cp?

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if file == copy:
+        return
+    with open(file, "r") as file_in, open(copy, "w") as file_out:
+        file_out.write(file_in.read())
```
YukihiroSM, line 6:
```suggestion
   if file != copy: 
          with open(file, "r") as file_in, open(copy, "w") as file_out:
                  file_out.write(file_in.read())
```

YukihiroSM (changes requested):

app/main.py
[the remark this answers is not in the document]
S-SMTN: Was it in the task/tests?

olena-hul: @S-SMTN This tasks requires manual testing. So please fix comments

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, file, copy = command.split(" ")
```
olena-hul, line 2: No need to specify space as a separator - it is a default argument for split

olena-hul (changes requested):

S-SMTN:
Please write the fully descriptive task. Students have to study the program developing, not the thought reading

olena-hul (approved):

PR #445, opened 2023-06-06
Copy file function implemented

thorium-zero:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+from shutil import copy
+
+
+def copy_file(command: str) -> None:
+    comm_data = command.split()
```
YuriiKindrat, line 5: you can unpack these variables in the easiest way
`command, old_file, new_file = command.split()` 
so there is won't be need to use indexes later

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,28 @@
-# write your code here
+from shutil import copy
+from os.path import exists
+
+
+class ArgumentException(Exception):
+    pass
```
Oleg-s-s, line 6: I think its not necessary to create a custom exception. You can use ValueError and then combine em in except block

app/main.py
```diff
@@ -1 +1,28 @@
...
+            raise ArgumentException("Destination must differ from the source")
+        if not exists(source):
+            raise FileNotFoundError(f"Unable to find the "
+                                    f"source file at '{source}'")
+    except ValueError:
+        print("You have to provide 3 parameters in the command string")
+    except ArgumentException as er:
+        print(f"Invalid arguments: {er}")
+    except FileNotFoundError as er:
+        print(er)
```
Oleg-s-s, line 24:
```suggestion
        if cmd != "cp":
            raise ValueError("Unknown command")
        if source == destination:
            raise ValueError("Destination must differ from the source")
        if not exists(source):
            raise FileNotFoundError(f"Unable to find the "
                                    f"source file at '{source}'")
    except (ValueError, FileNotFoundError) as error:
        print(f"Error: {error}")
```
Smth like this ^ What do you think?

Oleg-s-s (changes requested):

Oleg-s-s (approved):
:+1:

PR #448, opened 2023-06-07
task

n1ckmel:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split(" ")
+    file_name = command_parts[1]
```
ArsenPidhoretskyi, line 3: Validate command please

ArsenPidhoretskyi (changes requested):
https://github.com/n1ckmel/py-copy-file/blob/develop/checklist.md
Read about code efficency

vasylhnatiuk (changes requested):
validate if command have 3 arguments

app/main.py
```diff
@@ -2,7 +2,7 @@ def copy_file(command: str) -> None:
     command_parts = command.split(" ")
     file_name = command_parts[1]
     copy_file_name = command_parts[2]
-    if command_parts[0] != "cp" or len(command) != 3:
+    if command_parts[0] != "cp" or len(command_parts) != 3:
```
vasylhnatiuk, line 5: you must do it before` file_name = command_parts[1]`, because if command doesn't have 3 arguments code will be broken

vasylhnatiuk (approved):

PR #449, opened 2023-06-08
Solution

Andriy-Sydorenko:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+from os import path
+import sys
+
+
+def copy_file(command: str) -> str or None:
+    cmd, main_file, new_file = command.split()
```
thorium-zero, line 6: This operation is potentially dangerous and may lead to unhandled exception due to inability to split the command string if its formatted incorrectly. You should `try:` this line for `ValueError` too if you wish to keep your code safe and robust

app/main.py
```diff
@@ -1 +1,21 @@
...
+    source = path.join(sys.path[0], main_file)
+    destination = path.join(sys.path[0], new_file)
```
thorium-zero, line 8: there's a function `os.getcwd()` that would turn the current working directory, that, in some cases could be changed prior to the module where your `sys.path[0]` call is located. Also, that function would, arguably, improve your code readability

app/main.py
```diff
@@ -1 +1,21 @@
...
+        if main_file == new_file:
+            raise ValueError("Cannot copy file to itself!")
+
+    # we can use copy() from shutil, but I've made it in more "vanilla" way
+        with open(source, "r") as file_in, open(destination, "w") as file_out:
+            file_out.write(file_in.read())
+    except ValueError as e:
+        print(f"ValueError: {e}")
+    else:
+        print("File copied successfully!")
```
thorium-zero, line 21: you could move that print to the bottom of your `try:` block

app/main.py
```diff
@@ -1 +1,21 @@
...
+    cmd, main_file, new_file = command.split()
+    source = path.join(sys.path[0], main_file)
+    destination = path.join(sys.path[0], new_file)
+    try:
+        if cmd != "cp":
+            raise ValueError("Wrong command! You should use `cp`.")
+        if main_file == new_file:
+            raise ValueError("Cannot copy file to itself!")
+
+    # we can use copy() from shutil, but I've made it in more "vanilla" way
```
thorium-zero, line 15: If you have to leave a comment for your mentor, use your pull request comments, rather than code comments. Please, remove this comment block :)

thorium-zero (changes requested):

app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+from os import path, getcwd
+
+
+def copy_file(command: str) -> str or None:
+
+    try:
+        cmd, main_file, new_file = command.split()
+        source = fr"{getcwd()}\{main_file}"
+
```
thorium-zero, line 9: consider removing this line due to the fact, that we can work with relative paths here. theres no need to get our working dir at all, tbh, but still, it is better to use `getcwd` than `sys.path[0]` if you insist. Also, If you wish to keep you `source` path as an absolute one, you should use `os.path.join` to make it

thorium-zero (changes requested):

thorium-zero (changes requested):
If you wish to keep the source as an absolute path, update it's construction using `os.path.join`, or simply make so the program accept that source path as it is, unchanged. I'd prefer it that way, tbh

thorium-zero (approved):
GJ!🔥

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from os import path
+
+
+def copy_file(command: str) -> str or None:
```
YuriiKindrat, line 4: in which row the function will return a str ?)

YuriiKindrat (changes requested):

PR #450, opened 2023-06-09
Solution

YevheniiIovchu:


mariiahorbova (changes requested):

mariiahorbova (approved):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_command = command.split()
```
YuriiKindrat, line 2: could u make split like this?

app/main.py
```diff
@@ -1 +1,14 @@
...
+        new_file = old_file.split(".")[0] + "-copy.txt"
+        with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+            file_out.write(file_in.read())
+    if list_command[0] == "cp" and len(list_command) == 3:
+        old_file = list_command[1]
+        new_file = list_command[2]
+        if old_file == new_file:
+            return
+        with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+            file_out.write(file_in.read())
```
YuriiKindrat, line 14: wake ur code DRY here

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,28 @@
...
+    cmd = old_file = new_file = None
+    try:
+        cmd = command.split()[0]
+        old_file = command.split()[1]
+        new_file = command.split()[2]
+    except IndexError:
+        print(f"You don't put all necessary data"
+              f"{{'cmd': {cmd},"
+              f"'old_file': {old_file},"
+              f"'new_file': {new_file}}}")
```
Oleg-s-s, line 11: You can simplify this code by using `try` with `cmd, old_file, new_file = command.split()` and `ValueError`

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1 +1,25 @@
...
+        print("You put incorrect data")
+    else:
+        try:
+            if old_file == new_file:
+                raise ValueError(f"You can't copy {old_file}"
+                                 f" with the same name!")
+            if cmd != "cp":
+                raise ValueError(f"This command {cmd} unavailable."
+                                 f" You can just copy file!")
+            if not (old_file.endswith(".txt") and new_file.endswith(".txt")):
```
LiudmylaKulinchenko, line 14: Linux cp command works not only for txt files, so you could remove this check

app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        cmd, old_file, new_file = command.split()
+    except ValueError:
+        print("You put incorrect data")
+    else:
+        try:
+            if old_file == new_file:
```
LiudmylaKulinchenko, line 8: You can also skip raising an error here and just use reversed condition according to the task condition

app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        cmd, old_file, new_file = command.split()
+    except ValueError:
```
LiudmylaKulinchenko, line 4: You can avoid nested try blocks in your case
Skip handling ValueError here and move code from `else` on the upper level. The ValueError will be caught by `except ValueError as ve: print(f"{ve}")`

LiudmylaKulinchenko (changes requested):

app/main.py
LiudmylaKulinchenko: Also you can just add check if command len == 3

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, old_file, new_file = command.split()
+    if old_file == new_file:
+        return
```
Oleg-s-s, line 4: Why not simply using reverse condition here. That will exclude using empty return

Oleg-s-s (changes requested):

YuriiKindrat (approved):

PR #451, opened 2023-06-09
implemented py-copy-file

mariiahorbova:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import os
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+    source_file = parts[1]
+    destination_file = parts[2]
```
Andriy-Sydorenko, line 7: It's better to use `try` for the whole function, to catch some `ValueError` exceptions, e.g. if the user is trying to copy the file to itself. or if the input is invalid or isn't splittable to 3 elements (f.e. "cp smth_smth"), you need to raise an appropriate exception, then catch them in `except`, and print the error message.

Andriy-Sydorenko (changes requested):

Andriy-Sydorenko (approved):
Good job!))

Oleg-s-s (changes requested):
_From checklist:_
Make sure you check that the `cp` input command is correct.

Oleg-s-s:
You've already implemented this check, but deleted it with len() check

Oleg-s-s (approved):
:+1:

PR #452, opened 2023-06-09
solution

Tarasidze:


app/main.py
```diff
@@ -1 +1,19 @@
...
+        return
+
+    cmd = command_list[0]
+    file_name_in = command_list[1]
+    file_name_out = command_list[2]
+
+    if file_name_in == file_name_out or cmd != "cp":
+        return
+
+    with (open(file_name_in, "r") as file_in,
```
AntonShpakovych, line 14: Imagine situation when file doesn't exist or even more if u don't have permission to read this file

AntonShpakovych (changes requested):

app/main.py
AntonShpakovych, line 14: Imagine situation when file doesn't exist or even more if u don't have permission to read this file. Also mode by default "r" so u can delete this "r" if u want of course

AntonShpakovych (changes requested):

app/main.py
Tarasidze: good point, thanks

app/main.py
```diff
@@ -11,7 +11,12 @@ def copy_file(command: str) -> None:
     if file_name_in == file_name_out or cmd != "cp":
         return
 
-    with (open(file_name_in, "r") as file_in,
+    try:
+        open(file_name_in)
+    except (FileExistsError, PermissionError) as e:
```
AntonShpakovych, line 16: FileExistsError:
FileExistsError is raised when you try to create or rename a file or directory, but a file or directory with the same name already exists in the target location. It indicates that the operation failed because the file or directory already exists. This error is specific to situations where the file or directory you're trying to create or rename already exists.

For example, if you attempt to create a file with the name "example.txt" in a directory where another file with the same name already exists, a FileExistsError will be raised.

FileNotFoundError:
FileNotFoundError is raised when you try to access or perform operations on a file or directory that does not exist in the specified location. It indicates that the file or directory could not be found at the given path.

For example, if you try to open a file for reading with the path "nonexistent.txt," and there is no file with that name in the specified location, a FileNotFoundError will be raised.

app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
+
+    if len(command_list) < 3:
+        return
+
+    cmd = command_list[0]
```
AntonShpakovych, line 7: u use cmd only in condition so u don't need to unpack this value, just use if not startwith method.  And of course u can group up your condition using operator  (or)

app/main.py
Tarasidze: Nice explanation!

AntonShpakovych (approved):

app/main.py
```diff
@@ -1 +1,24 @@
...
+    file_name_in = command_list[1]
+    file_name_out = command_list[2]
+
+    if file_name_in == file_name_out or cmd != "cp":
+        return
+
+    try:
+        open(file_name_in)
```
AntonShpakovych, line 15: I think we shouldn't use try except for memoization because try except use too much memory. But maybe a mentor will help us with this later

app/main.py
AntonShpakovych, line 7: u use cmd only in condition so u don't need to unpack this value, just use if not startwith method. And of course u can group up your condition using operator (or)

AntonShpakovych (changes requested):
change cmd and i will approve this

app/main.py
Tarasidze: Most likely, it depends on the task to be performed and who will use your application

AntonShpakovych (approved):
Good job!

app/main.py
```diff
@@ -1 +1,23 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
```

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+    if len(command_list) < 3 or not command.startswith("cp"):
+        return
+
+    file_name_in = command_list[1]
+    file_name_out = command_list[2]
```
YuriiKindrat, line 8: could u unpack this like this 
`arg1, arg2, arg3 = command_list`

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+    if file_name_in == file_name_out:
+        return
```
YuriiKindrat, line 11: u can use backward condition here

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+    try:
+        open(file_name_in)
+    except (FileNotFoundError, PermissionError) as e:
```
YuriiKindrat, line 15: why do u use this try-except block separately from CM ?

app/main.py
```diff
@@ -1 +1,23 @@
...
+        raise f"Error {e}. The file does not exist or can not be read"
+
+    with (open(file_name_in) as file_in,
+          open(file_name_out, "w") as file_out):
+        file_content = file_in.read()
+        file_out.write(file_content)
+
+    return
```
YuriiKindrat, line 23: why do u need this ?)

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+    try:
+        open(file_name_in)
+    except (FileNotFoundError, PermissionError) as e:
+        raise f"Error {e}. The file does not exist or can not be read"
+
+    with (open(file_name_in) as file_in,
+          open(file_name_out, "w") as file_out):
+        file_content = file_in.read()
+        file_out.write(file_content)
```
YuriiKindrat, line 21: could u make this in one line ?

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+
+    if len(command_list) == 3 or command.startswith("cp "):
```
Oleg-s-s, line 4: `or` ?

app/main.py
```diff
@@ -1 +1,13 @@
...
+        _, file_name_in, file_name_out = command_list
+
+        if file_name_in != file_name_out:
+            try:
+                with (open(file_name_in) as file_in,
+                      open(file_name_out, "w") as file_out):
+                    file_out.write(file_in.read())
+            except (FileNotFoundError, PermissionError) as e:
+                raise f"Error {e}. The file does not exist or can not be read"
```
Oleg-s-s, line 13: `{e}` will already contain the error info, no need in additional `The file does not exist or can not be read` that can only confuse the user

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if len(command_list) == 3 and command.startswith("cp "):
+        _, file_name_in, file_name_out = command_list
+
+        if file_name_in != file_name_out:
+            try:
+                with (open(file_name_in) as file_in,
+                      open(file_name_out, "w") as file_out):
+                    file_out.write(file_in.read())
+            except (FileNotFoundError, PermissionError) as e:
+                raise e
```
Oleg-s-s, line 13: Now lets think, do you actually need this try block if we are not handling exceptions at all, and just calling em?))

Oleg-s-s (changes requested):

app/main.py
LiudmylaKulinchenko: It's better to use str(e). It is not guaranteed across all exception types that the exception will have message attribute

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job!! 🍬🎈

PR #453, opened 2023-06-09
Solution

diana-shyrokikh:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
+    file_to_copy = command[1]
+    new_file = command[-1]
```
Tarasidze, line 4: At least we need to be sure that "cp" exists on our "command" line.

app/main.py
```diff
@@ -1 +1,9 @@
...
+
+    if file_to_copy != new_file:
+        with open(file_to_copy, "r") as source, open(new_file, "w") as copy:
```
Tarasidze, line 7: Also, I think it would be great to check if our file exists. Maybe use an exception.

Tarasidze, line 7: Also, I think it would be great to check if our file exists. Maybe use an exception.

app/main.py
Tarasidze, line 4: At least we need to be sure that "cp" is in our "command" line.

app/main.py
Tarasidze, line 7: Also, I think it would be great to check if our file exists. Maybe use an exception.

Tarasidze (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
```
oleksiikolii, line 2: You can use `command.split()` here, `" "` is given by default

app/main.py
oleksiikolii, line 4: You don't follow your logic here. If you know that command has definitely 3 items just pick `command[2]` here and if it doesn't work that means given command wasn't valid and you need to raise some error here.  What if `command` was `"cp path1.txt some_random_text path2.txt some_random_text"`?

app/main.py
```diff
@@ -1 +1,9 @@
...
+    file_to_copy = command[1]
+    new_file = command[-1]
+
+    if file_to_copy != new_file:
+        with open(file_to_copy, "r") as source, open(new_file, "w") as copy:
+            for line in source:
+                copy.write(line)
```
oleksiikolii, line 9: You can copy file entirely no need to copy line by line, don't use `for` loop here

oleksiikolii (changes requested):
Add validation to the `command`

Tarasidze (approved):
Nice

app/main.py
```diff
@@ -1 +1,31 @@
...
+
+    if len(command) != 3:
+        raise CopyCommandSyntaxError(
+            "Invalid copy command syntax.\n"
+            "For a successful copy "
+            "the command should look like this: "
+            "cp file.txt file-copy.txt"
+        )
+
+    file_to_copy = command[1]
```
YuriiKindrat, line 22: u could unpack ur command like this
`var1, var2, var3 = command` after len checking

LiudmylaKulinchenko (approved):
Well done! 🌼

PR #454, opened 2023-06-10
Solution

AntonShpakovych:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> bool:
+    if not command.startswith("cp") or len(command.split(" ")) != 3:
```
ant-komarov, line 2: What if command will start with "cpu..."?

ant-komarov (changes requested):

app/main.py
AntonShpakovych: But give me example where cpu command can be given in function wich works with file

ant-komarov (approved):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> bool:
+    try:
+        command, file, new_file = command.split(" ")
```
YuriiKindrat, line 3: no need to use space inside split method, it uses space by default

app/main.py
```diff
@@ -1 +1,12 @@
...
+        if file != new_file and command == "cp":
+            with open(file) as file, open(new_file, "a+") as new_file:
+                new_file.write(file.read())
+        else:
+            return False
+    except (ValueError, FileNotFoundError, PermissionError):
+        return False
+    else:
+        return True
```
YuriiKindrat, line 12: why do u need this returns here? it seems to me, that this func shouldn't return smth

YuriiKindrat (changes requested):

app/main.py
AntonShpakovych: Yes, I know it.  Okey I will change this, but I think it depends on people. I prefer use with quotes

app/main.py
AntonShpakovych: In documentation for this function we don't have info about return or print smth. So I created this function like I wont

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        current_command, file, new_file = command.split()
+        if file != new_file and current_command == "cp":
+            with open(file) as file, open(new_file, "a+") as new_file:
+                new_file.write(file.read())
+    except (ValueError, FileNotFoundError, PermissionError):
+        raise
```
YuriiKindrat, line 8: gj, but why do u need to catch error to reraise them ?)

YuriiKindrat (changes requested):

app/main.py
AntonShpakovych: Okay, I fixed this => To return string with message. But we have moment when command "cp" not found in our function parameter, so for this code without return False, what the user should think or must do?

YuriiKindrat: if we talk about function, i think it is better to return string with message, or u can raise an error with message, or smth like this 
it seems to me, that False returning is not enough descriptive for user to understand, that command is incorrect )

YuriiKindrat (approved):

PR #455, opened 2023-06-10
solution

oleksiikolii:


app/main.py
```diff
@@ -1 +1,31 @@
...
+        if command[0] == "cp":
+            if command[1] != command[2]:
+                with (open(command[1], "r") as source,
+                      open(command[2], "w") as new_file):
+                    new_file.write(source.read())
+                print("Copying done!")
+                return
+
+            raise PathError(f"File path {command[1]} and "
+                            f"'{command[2]}' can not be the same!")
```
diana-shyrokikh, line 23: from README: 
"It must do nothing in case the user is trying to copy file to file with the same name"
So you can delete this error raise and do nothing with it

diana-shyrokikh (changes requested):

diana-shyrokikh (approved):

app/main.py
```diff
@@ -1 +1,25 @@
...
+                      open(command[2], "w") as new_file):
+                    new_file.write(source.read())
+                print("Copying done!")
+                return
+
+        raise CommandError(f"Unknown command: {command[0]}\n"
+                           "Did you mean 'cp'?")
+
+    except CommandError as e:
+        print(e)
```
YuriiKindrat, line 23: why do u need to raise exception and then catch it ?

app/main.py
```diff
@@ -1 +1,25 @@
...
+
+
+def copy_file(command: str) -> None:
+    try:
+        command = command.split()
+        if len(command) != 3:
+            raise CommandError("Unsupported format of command")
+
+        if command[0] == "cp":
+            if command[1] != command[2]:
```
YuriiKindrat, line 12: could u unite this ?)

app/main.py
```diff
@@ -1 +1,25 @@
...
+                with (open(command[1], "r") as source,
+                      open(command[2], "w") as new_file):
+                    new_file.write(source.read())
+                print("Copying done!")
+                return
```
YuriiKindrat, line 17: why do u need blank return here?

app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+class CommandError(Exception):
+    pass
+
+
+def copy_file(command: str) -> None:
+    try:
+        command = command.split()
+        if len(command) != 3:
+            raise CommandError("Unsupported format of command")
```
YuriiKindrat, line 9: could u unpack ur var after this check ? 
`arg1, arg2, arg3 = command_list`

YuriiKindrat: to avoid using indexes and increase readability

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+        cmd, file_name, new_file_name = command
+        if cmd == "cp" and file_name != new_file_name:
+            with (open(file_name, "r") as source,
+                    open(new_file_name, "w") as new_file):
+                new_file.write(source.read())
+            print("Copying done!")
+
+        raise CommandError(f"Unknown command: {command[0]}\n"
```
Oleg-s-s, line 18: You need to change this, error is raising both on `cmd == "cp"` this check and on `file_name != new_file_name` this check and even if code finish correctly

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+        cmd, file_name, new_file_name = command
+        if cmd == "cp" and file_name != new_file_name:
+            with (open(file_name, "r") as source,
+                    open(new_file_name, "w") as new_file):
+                new_file.write(source.read())
+            print("Copying done!")
+        else:
+            raise CommandError(f"Unknown command: {cmd}\n"
+                               "Did you mean 'cp'?")
```
Oleg-s-s, line 19: Again you are handling only `!=cp` here, what about `file == new_one`

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    except FileNotFoundError as e:
+        print(e)
```
Oleg-s-s, line 22: Why not placing this try/except block on context manager directly, that would improve readability for sure

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1 +1,24 @@
...
+
+
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3:
+        raise CommandError("Unsupported format of command")
+
+    cmd, file_name, new_file_name = command
+    if cmd == "cp":
+        if file_name != new_file_name:
```
LiudmylaKulinchenko, line 12: You can combine the two if conditions into a single line using the logical operator

app/main.py
```diff
@@ -1 +1,24 @@
...
+            try:
+                with (open(file_name, "r") as source,
+                        open(new_file_name, "w") as new_file):
+                    new_file.write(source.read())
+                print("Copying done!")
+            except FileNotFoundError as e:
+                print(e)
+        else:
+            print("File with the same name cannot be copied!")
```
LiudmylaKulinchenko, line 21: The function `must do nothing in case the user is trying to copy file to file with the same name` according to the task condition, you should not print anything

LiudmylaKulinchenko (changes requested):

app/main.py
oleksiikolii: Oleg just said that I need to track both conditions and raise different exceptions

LiudmylaKulinchenko (approved):
Well done! 🍬

PR #456, opened 2023-06-11
'Solution'

ant-komarov:


erikagayan (approved):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None | str:
+    if not command.startswith("cp ") or len(command.split(" ")) != 3:
+        return "Entered incorrect command!!!"
+    old_file_name = command.split(" ")[1]
+    new_file_name = command.split(" ")[2]
```
YuriiKindrat, line 5: make ur code DRY, u are using split for 3 times 
also no need to use space inside split method, it uses space by default
also u could unpack ur var like this `var1, var2, var3 = command.split()`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
...
+            try:
+                with (open(old_file_name, "r") as old_file,
+                      open(new_file_name, "w") as new_file):
+                    new_file.write(old_file.read())
+            except FileNotFoundError:
+                return f"There is no file with name {old_file_name}"
+            except PermissionError:
+                return f"You don't have permission to file {old_file_name}"
+    except ValueError as error:
+        print(error)
```
Oleg-s-s, line 16: this part handle only this line
`cmd_name, old_file_name, new_file_name = command.split()`
so I'm thinking about placing except after it to improve readability(don't forget to stop the function)

Oleg-s-s (changes requested):

LiudmylaKulinchenko (approved):
Well done ☀

PR #458, opened 2023-06-11
Copy file (First commit)

IhorVoskoboinikov:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    com, first_file, second_file = command.split(" ")
+    if first_file == second_file:
+        return
```
volodymyr-komarnyckyi, line 4: Add print here so you can see why the program isn't executing

volodymyr-komarnyckyi (changes requested):

app/main.py
```diff
@@ -1,7 +1,10 @@
 def copy_file(command: str) -> None:
-    com, first_file, second_file = command.split(" ")
+    com, first_file, second_file = command.split("")
```
volodymyr-komarnyckyi, line 2: no need to use quotes inside split method, just use split()

volodymyr-komarnyckyi (changes requested):

volodymyr-komarnyckyi (commented):
Great job

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    com, first_file, second_file = command.split()
```
LiudmylaKulinchenko, line 2: "Make sure you check that the cp input command is correct." it is from the **checklist**

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if first_file != second_file:
+        with (open(first_file, "r") as file_in,
+              open(second_file, "w") as file_out):
+            data = file_in.read()
```
LiudmylaKulinchenko, line 6: You can skip creating a new variable `data` here and put .read() directly in the write()

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
...
+def copy_file(command: str) -> None:
+    com, first_file, second_file = command.split()
+
+    if first_file != second_file and com == "cp":
+        with (open(first_file, "r") as file_in,
+              open(second_file, "w") as file_out):
+            file_out.write(file_in.read())
+            print(f"The data was successfully copied from the "
+                  f"{first_file} to a {second_file}")
+    print("Copy error! The command should be 'cp' "
```
LiudmylaKulinchenko, line 10: You print this anyway even if no errors occurred. You can just skip it if you want according to the task condition

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    com, first_file, second_file = command.split()
```
LiudmylaKulinchenko, line 2: I'd suggest using the whole word if you aren't completely sure it's necessary. It only saves you few characters anyway

LiudmylaKulinchenko (approved):
Good job 😎☀

PR #459, opened 2023-06-11
Solution #1

erikagayan:


IhorVoskoboinikov (approved):
Great job!

For better readability, the following edits could be added:
com, first_file, second_file = command.split(" ")

And as suggested in the assignment to use multiple file opening
with open(first_file, "r") as file_in, open(second_file, "w") as file_out:

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    words = command.split(" ")
```
also u can unpack this like this, to avoid using indexes and increase a readability 

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+    if words[1] == words[2]:
+        return
```
YuriiKindrat, line 5: could u use backward condition here?

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+    with open(words[1], "r") as old_file:
+        with open(words[2], "w") as new_file:
+            content = old_file.read()
+            new_file.write(content)
```
YuriiKindrat, line 10: image
fix this

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, first_file, second_file = command.split(" ")
```

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+    if first_file != second_file:
+        with open(first_file, "r") as old_file, \
+                open(second_file, "w") as new_file:
```
YuriiKindrat, line 6: try to avoid using linebreakers in ur code

app/main.py
```diff
@@ -1 +1,10 @@
...
+            content = old_file.read()
+            new_file.write(content)
+    else:
+        return
```
YuriiKindrat, line 10: why do u need this ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #461, opened 2023-06-11
Solution

Po-SvinSkipodla:


app/main.py
```diff
@@ -1,7 +1,8 @@
 def copy_file(command: str) -> None:
     command_ls = command.split()
```
Altahoma, line 2: You can probaly simplify code using assigning multiple variables in one line.

app/main.py
```diff
@@ -1,7 +1,8 @@
...
     try:
-        with open(command_ls[1], "r") as file_in, open(command_ls[2], "x") as file_out:
+        with (open(command_ls[1], "r") as file_in,
+              open(command_ls[2], "x") as file_out):
             file_out.write(file_in.read())
     except FileExistsError:
         pass
```
Altahoma, line 8: Try to avoid passing in except blocks.

Altahoma (changes requested):

app/main.py
Po-SvinSkipodla: You mean `cmd, file_in, file_out = command.split()`?

app/main.py
Po-SvinSkipodla: I'm not quite understand what you want me to do, cause we need to do nothing if we try to create file with the same name, so we need try except block. If you know another solution, give me a tip, so i can figure it out on my own.

app/main.py
Altahoma: Yah, further we will have easy access to these variables

app/main.py
Altahoma: 'passing in except' can lead to silent errors and make it difficult to identify and handle exceptions properly. Error handling means processing them in a certain way.

Altahoma (approved):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, file_in, file_out = command.split()
```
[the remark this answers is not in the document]
LiudmylaKulinchenko: You didn't make any changes 😮

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if not cmd == "cp":
```
Oleg-s-s, line 3: It will be better to do it like:
```suggestion
    if cmd == "cp":
        ...
```
There will be no need to use additional return then

Oleg-s-s (changes requested):

LiudmylaKulinchenko (approved):
Well done! 🌼

PR #463, opened 2023-06-12
Solution

volodymyr-komarnyckyi:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        print("Invalid command!")
+        return
```
Oleg-s-s, line 5: Why not just return print?

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+    if parts[1] == parts[2]:
+        print("The source file and destination file have the same name!")
+        return
```
Oleg-s-s, line 9: Same here

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
Oleg-s-s, line 2: You can make something like `cmd, old_file, new_file = parts` after len() check to improve code readability

app/main.py
```diff
@@ -1 +1,19 @@
...
+        print("Invalid command!")
+        return
+
+    if parts[1] == parts[2]:
+        print("The source file and destination file have the same name!")
+        return
+
+    try:
+        with open(parts[1], "r") as source_file, \
+                open(parts[2], "w") as dest_file:
```
Oleg-s-s, line 13: Avoid using linebreaks

Oleg-s-s (changes requested):

Oleg-s-s (approved):
:+1:

PR #464, opened 2023-06-12
'Solution'

yakhlamova:


YKonoplyov (approved):
Nice work👍

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy(command: str) -> None:
+    command = command.split()
+    if command[1] == command[2] or command[0] != "cp":
+        return
```
Oleg-s-s, line 4: Why you left empty return? Provide at least either `print` or some `error` message.

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy(command: str) -> None:
+    command = command.split()
```
Oleg-s-s, line 2: You can split using a bunch of variables for better readability, f.e.
`cmd, file_name, new_file_name = command.split()`

Oleg-s-s (changes requested):

app/main.py
yakhlamova: Ok, I just thought it would be better. In the task written, it must do nothing, and Danylo in the lecture said it`s ok to use bare return.

Oleg-s-s: In that case it would be better to use reverse logic and complete copy if
`command[1] != command[2] and command[0] == "cp"`
don't you agree?)

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy(command: str) -> None:
+    cmd, file_name, new_file_name = command.split()
+    if file_name == new_file_name or cmd != "cp":
+        print("Does nothing")
```
Oleg-s-s, line 4: Your function don't stop here. You can take in mind my higher comment and do that way or at least make this version work :upside_down_face:

Oleg-s-s (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy(command: str) -> None:
+    cmd, current, future = command.split()
```
LiudmylaKulinchenko, line 2: It still will be much more descriptive to name variables like `current_file`

LiudmylaKulinchenko (changes requested):

app/main.py
LiudmylaKulinchenko: @yakhlamova I meant the same for the `future` variable. You should name it more descriptively

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy(command: str) -> None:
+    cmd, current_file, future = command.split()
```
LiudmylaKulinchenko, line 2: Use more descriptive name for the `future` variable, please

LiudmylaKulinchenko (changes requested):

app/main.py
yakhlamova: I tried, but then I have a problem with the line too long, so I was in search of a nice name for this variable but I can`t pick up so I left what it was

Oleg-s-s: `cmd, current_file, future_file = command.split()`
`with (open(current_file, "r") as file_in,
              open(future_file, "w") as file_out):`

app/main.py
Oleg-s-s, line 2:
```suggestion
    cmd, current_file, future_file = command.split()
```

app/main.py
```diff
@@ -1 +1,5 @@
...
+    if current_file != future or cmd == "cp":
+        with open(current_file, "r") as file_in, open(future, "w") as file_out:
```
Oleg-s-s, line 4: Dont forget that you can cover everything in brackets (almost)
```suggestion
        with (open(current_file, "r") as file_in,
              open(future_file, "w") as file_out):
```

LiudmylaKulinchenko (approved):

LiudmylaKulinchenko (approved):
Well done!! 🍬🌺

PR #466, opened 2023-06-12
solution

Gvilion:


OlhaTryhub (changes requested):

OlhaTryhub (approved):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    copied_file = command.split()[1]
+    new_file = command.split()[2]
```
LiudmylaKulinchenko, line 3: Make it one line, use unpacking for `command`

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    copied_file, new_file = command.split()[1:]
```
YuriiKindrat, line 2: if u don't need some value while unpacking it is more popular way to use `_`
f.e. `_, var1, var2 = command.split()`

app/main.py
```diff
@@ -1 +1,7 @@
...
+
+    if copied_file != new_file:
+        with (open(copied_file, "r") as c_file,
+              open(new_file, "w") as n_file):
+            n_file.write(c_file.read())
```
YuriiKindrat, line 7: c_file and n_file is a bad names for vars )

YuriiKindrat (changes requested):

YuriiKindrat (approved):

PR #467, opened 2023-06-12
Solution

Kirontiko:


Mr-Furman (approved):
Man, your code is amazing! It even has code for calling functions with short instructions. GJ! See you at Google.

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    target_file, new_file = command.split()[1:]
```
LiudmylaKulinchenko, line 2: "Make sure you check that the cp input command is correct." it is from the **checklist**

app/main.py
```diff
@@ -1 +1,14 @@
...
+    if target_file == new_file:
+        return
```
LiudmylaKulinchenko, line 4: try not to use blank return here, make a backward condition

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, target_file, new_file = command.split()
+    if cmd != "cp":
+        raise ValueError("Command not found.")
+    if not target_file.endswith(".txt") or not new_file.endswith(".txt"):
```
LiudmylaKulinchenko, line 5: The cp command in Linux is not limited to working only with text files

app/main.py
```diff
@@ -1 +1,19 @@
...
+        raise FileNotFoundError("File is not found")
+
+    if target_file != new_file:
+
```
LiudmylaKulinchenko, line 9: You can remove an empty line here to keep the same code style in the file

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, target_file, new_file = command.split()
+    if cmd != "cp":
+        raise ValueError("Command not found.")
+    if not target_file.endswith(".txt") or not new_file.endswith(".txt"):
+        raise FileNotFoundError("File is not found")
```
LiudmylaKulinchenko, line 6: Are u sure the error is appropriate for a check if files are text files?

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Well done! 😋🎈

PR #468, opened 2023-06-12
My solution

kovaliskoveronika:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[1] != command[2]:
```
LiudmylaKulinchenko, line 3: Use unpacking instead of indexing command multiple times

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, origin, copy = command.split()
+    if origin != copy:
```
LiudmylaKulinchenko, line 3: Add, please, check that the cp input command is correct, as mentioned in the **checklist**

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, origin, copy = command.split()
+    if command != "cp" or ".txt" in (origin, copy):
+        print("Wrong command")
+        return
```
Oleg-s-s, line 5: According to task you should do nothing in that case, so you can use reverse condition and don't use additional return)

Oleg-s-s (changes requested):

Oleg-s-s (approved):
:+1:

PR #469, opened 2023-06-12
Solution

YKonoplyov:


app/copy.txt
Po-SvinSkipodla, line 1: You don't need this file)

app/orig.txt
Po-SvinSkipodla, line 1: You don't need this file)

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split(" ")) < 3:
```
Po-SvinSkipodla, line 2: You can write just `command.split()`

app/main.py
```diff
@@ -1 +1,11 @@
...
+        return
+
+    command_list = command.split(" ")
+    original_file = command_list[1]
+    file_to_copy = command_list[2]
```
Po-SvinSkipodla, line 7: You can assign 3 variables in 1 line and give it exact name

app/main.py
```diff
@@ -1 +1,11 @@
...
+
+    with open(original_file, "r") as orig, open(file_to_copy, "a") as copy:
+        for line in orig.readlines():
+            copy.write(line)
```
Po-SvinSkipodla, line 11: Use method read() instead of readlines()

app/main.py
Po-SvinSkipodla, line 1: Where is check for the same names for orig and copy?

Po-SvinSkipodla (changes requested):

app/main.py
```diff
@@ -1,11 +1,8 @@
 def copy_file(command: str) -> None:
-    if len(command.split(" ")) < 3:
-        return
+    command, original_file, file_to_copy = command.split()
 
-    command_list = command.split(" ")
-    original_file = command_list[1]
-    file_to_copy = command_list[2]
+    if len(command.split()) < 3 or original_file == file_to_copy:
+        return
```
Po-SvinSkipodla, line 5: You can also add command check for cp command

Po-SvinSkipodla (approved):

app/copy.txt
LiudmylaKulinchenko, line 1: Remove unnecessary files from github

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, original_file, file_to_copy = command.split()
+
+    if (len(command.split()) < 3
+            or original_file == file_to_copy
+            or command != "cp"):
+        return
```
LiudmylaKulinchenko, line 7: try not to use blank return here, make a backward condition

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, original_file, file_to_copy = command.split()
+
+    if not (len(command.split()) < 3
+            or original_file == file_to_copy
+            or command != "cp"):
+
+        with open(original_file, "r") as orig, open(file_to_copy, "a") as copy:
```
LiudmylaKulinchenko, line 8: I'd suggest using the whole word if you aren't completely sure it's necessary. Use brackets if you need to split in lines

LiudmylaKulinchenko (approved):
Good job! 😮🎈

PR #470, opened 2023-06-12
Solution

Kasatello:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_part = command.split()
+    cp, source, target = command_part
```
YuriiKindrat, line 3: could u make this in one line?

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if source == target:
+        print("Source and target are the same")
```
YuriiKindrat, line 5: why do u need this, if program will continue to work ? 
image

YuriiKindrat (changes requested):

PR #471, opened 2023-06-12
Solution for Copy File

OlhaTryhub:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
```

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if (len(command_list) == 3
+            and command_list[1] != command_list[2]
+            and command_list[0] == "cp"):
```
YuriiKindrat, line 5: couldu unpack ur vars like this
to avoid using indexes and increase readability

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Good job! 🚀

PR #472, opened 2023-06-12
my_commit

Mr-Furman:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if not command.startswith("cp"):
+        raise Exception("this interface support only 'cp' command")
+    ls_with_name = command.split()[1:]
+    if ls_with_name[0] != ls_with_name[1]:
+        with open(ls_with_name[0], "r") as file1,\
+                open(ls_with_name[1], "w") as file2:
```
YuriiKindrat, line 7: avoid using linebreakers in ur code

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if not command.startswith("cp"):
+        raise Exception("this interface support only 'cp' command")
+    ls_with_name = command.split()[1:]
```
YuriiKindrat, line 4: could u unpack like this 
`arg1, arg2, arg3 = command.split()` to avoid using indexes for better readability ?
also u can use underscores for useless values
f.e. `_, arg2, arg3 = command.split()`

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, source_file, res_file = command.split()
+    if not command.startswith("cp"):
+        raise Exception("this interface support only 'cp' command")
```
Oleg-s-s, line 4: It would be better to use more specific exception f.e. `ValueError`

Oleg-s-s (approved):
:+1:

PR #480, opened 2023-06-22
Solution

prykhodkom:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_words = command.split()
```
YukihiroSM, line 2: User can enter commands like: 
`cp file1.txt` - not enough parameters
`mv file1.txt file2.txt` - wrong command
`cp file1.txt file2.txt file3.txt` - too much parameters

Add something to check these cases as well.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) == 3:
+        if command[0] == "cp" and command[1] != command[2]:
+            with open(command[1], "r") as file_in, \
+                    open(command[2], "w") as file_out:
+                content_file = file_in.read()
+                file_out.write(content_file)
```
Bohdan-Salamakha, line 8: 1
what are command[1], command[2]?
create separate variables for them and call on demand
2
use parentheses instead of backslash to wrap the line
3
unite 7-8 lines into one line

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #481, opened 2023-06-23
Solution

NazarLytvyniuk:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command:
+        split_command = command.split(" ")
```
Polyakiv-Andrey, line 3: a space is set by default in the split method; you do not need to put it

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+        if len(split_command) == 3 and split_command[0] == "cp":
```
Polyakiv-Andrey, line 5: add here the third check through and, but check the files for inequality !=

app/main.py
```diff
@@ -1 +1,14 @@
...
+            file_name1 = split_command[1]
+            file_name2 = split_command[2]
+
+            if file_name1 == file_name2:
+                return
```
Polyakiv-Andrey, line 10: there should be no return in this function

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+            with open(file_name1, "r") as f1, open(file_name2, "w") as f2:
+                for line in f1.readlines():
```
Polyakiv-Andrey, line 13: you can avoid the loop using the read() method

app/main.py
```diff
@@ -1 +1,14 @@
...
+        split_command = command.split(" ")
+
+        if len(split_command) == 3 and split_command[0] == "cp":
+            file_name1 = split_command[1]
+            file_name2 = split_command[2]
+
+            if file_name1 == file_name2:
+                return
+
+            with open(file_name1, "r") as f1, open(file_name2, "w") as f2:
```
Polyakiv-Andrey, line 12: since you are using 2 files, it is better to rename them, so that it would be clear from which information it is written out and into which it is written

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #482, opened 2023-06-26
Function works with copying file content. Tested with .txt.

stepband:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, source_file, dest_file = command.split()
```
vsmutok, line 2: Validate command length before unpacking

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if source_file == dest_file:
+        return
```
vsmutok, line 4: also check if first command is `cp`

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #483, opened 2023-06-29
'Solution'

olmatiash:
Hello!
This func create new file with another name.

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3:
+        raise IndexError("list index out of range")
+    source_file = command[1]
+    destination_file = command[2]
+    if source_file == destination_file:
+        return
+    if command[0] == "cp":
+        with (open(source_file, "r") as f,
+                open(destination_file, "w") as s):
+            s.write(f.read())
+
+
+copy_file("cp file.txt new_file.txt")


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3:
+        raise IndexError("list index out of range")
+    source_file = command[1]
+    destination_file = command[2]
```
```suggestion
    cmd, source_file, destination_file = command
```
Use unzipping

app/main.py
```diff
@@ -1 +1,15 @@
...
+    if source_file == destination_file:
```
```suggestion
    if source_file == destination_file or cmd == "cp":
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+        return
+    if command[0] == "cp":
+        with (open(source_file, "r") as f,
+                open(destination_file, "w") as s):
+            s.write(f.read())
```
ArsenPidhoretskyi, line 12:
```suggestion
        with (
            open(source_file, "r") as source,
            open(destination_file, "w") as destination
        ):
            destination.write(source.read())
```
No need to name as one letter

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3:
+        raise IndexError("list index out of range")
```
vasylhnatiuk, line 4: why do you repeat error? it would be better to explain for user that command is wrong

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3:
+        raise IndexError("Your command is wrong")
+    cmd, source_file, destination_file = command
+    if source_file == destination_file or cmd == "cp":
+        return
+    if command[0] == "cp":
```
ArsenPidhoretskyi, line 8: This check becomes redundant

ArsenPidhoretskyi (changes requested):

Grayder0152 (approved):

PR #484, opened 2023-06-29
Solution

alecilchenko:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
```
Polyakiv-Andrey, line 2: the split method has a space by default

app/main.py
```diff
@@ -1 +1,9 @@
...
+    command_order = command[0]
+    origin_file = command[1]
+    copy_file = command[2]
+    if origin_file == copy_file or command_order != "cp":
+        return None
```
Polyakiv-Andrey, line 7: rewrite the code so that there would be no return at all

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #487, opened 2023-07-03
Solution

Veinmax:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_origin = command.split()[1]
```
vasylhnatiuk, line 2: why do you use split 3 times? and add check if command doesn't have 3 arguments

vasylhnatiuk (changes requested):

PR #488, opened 2023-07-03
Solution

Quiet-Klirik:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
```
Polyakiv-Andrey, line 2: split method is using space as default

app/main.py
```diff
@@ -1 +1,9 @@
...
+    source_file_name = command[1]
+    new_file_name = command[2]
+
+    if source_file_name != new_file_name:
+        with open(source_file_name, "r") as source_file, \
```
Polyakiv-Andrey, line 7: use brackets instead of /

Polyakiv-Andrey (changes requested):

Dimosphen1 (approved):
LGTM, nice work!

PR #489, opened 2023-07-04
copy file

FadieievaA:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd_list = command.split()
+    if cmd_list[1] != cmd_list[2]:
```
YukihiroSM, line 3: Use unpacking instead of getting elements from the list by index.

app/main.py
```diff
@@ -1 +1,6 @@
...
+        with open(cmd_list[1], "r") as file_in, \
+                open(cmd_list[2], "w") as file_out:
```
YukihiroSM, line 5: Use `()` instead of `\`, please.

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd_list = command.split()
```
YukihiroSM, line 2: Add a check, that command is `cp` and not something else.

YukihiroSM (changes requested):

YukihiroSM (approved):
Nice job

PR #492, opened 2023-07-05
Solution

BogdanFSD:
Please check

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        raise ValueError("Invalid command format")
+
+    cp, source_file, destination_file = parts
+
+    if source_file != destination_file:
+        with open(source_file, "r") as file_in, \
+                open(destination_file, "w") as file_out:
+            file_out.write(file_in.read())


app/main.py
```diff
@@ -1 +1,11 @@
...
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        raise ValueError("Invalid command format")
+
+    cp, source_file, destination_file = parts
+
+    if source_file != destination_file:
+        with open(source_file, "r") as file_in, \
+                open(destination_file, "w") as file_out:
```
ArsenPidhoretskyi, line 10:
```suggestion
        with (
            open(source_file_path, "r") as source_file,
            open(destination_file_path, "w") as destination_file
       ):
```

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        raise ValueError("Invalid command format")
+
+    cp, source_file, destination_file = parts
```
ArsenPidhoretskyi, line 6: source_file, destination_file is not true naming
it is path. Use source_file_path or source_path
The same with destination

ArsenPidhoretskyi (changes requested):
https://github.com/BogdanFSD/py-copy-file/blob/develop/checklist.md
image

If someone enter:
"mv file.txt file2.txt"
Add Check that cp == "cp"

app/main.py
BogdanFSD: I see
will upload changes now

vasylhnatiuk (approved):

PR #495, opened 2023-07-06
cp_like_Linux

CeO-PsH:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd_1 = command.split()
+    if cmd_1[1] != cmd_1[2]:
+        with open(cmd_1[1], "r") as file_in, open(cmd_1[2], "w") as file_out:
```
Bohdan-Salamakha, line 4: what are cmd_1[1], cmd_2[1]?
create separate variables and call them on demand

Bohdan-Salamakha (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd_1 = command.split()
+    name_of_old_file = cmd_1[1]
+    name_of_new_file = cmd_1[2]
+    if name_of_old_file != name_of_new_file:
+        with open(name_of_old_file, "r") \
+                as file_in, open(name_of_new_file, "w") as file_out:
```
Bohdan-Salamakha, line 7: use parentheses instead of backslash to wrap the line

Bohdan-Salamakha (changes requested):

app/main.py
CeO-PsH: @Bohdan-Salamakha This Features works only by python 3.10, and i can`t pass flake8 test when use ()

Bohdan-Salamakha (approved):

PR #496, opened 2023-07-07
'Solution'

govorunenko:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    with open(command.split()[1] + ".txt", "r") as text_file,\
```
olena-hul, line 2: Save split result to a separate variable

Validate that the command length is equal exactly to 3 arguments

Validate that command name is exactly "cp"

Use unpacking to access command parts

olena-hul (changes requested):

app/main.py
govorunenko: Done.

app/main.py
```diff
@@ -1 +1,12 @@
...
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+
+    if len(command_parts) != 3 or command_parts[0] != "cp":
+        return
+
+    _, source_name, copy_name = command_parts
+
+    with open(source_name, "r") as source_file,\
+            open(copy_name, "w") as copy_file:
```
Bohdan-Salamakha, line 10: use parentheses instead of backslash to wrap the line

app/main.py
```diff
@@ -1 +1,12 @@
...
+        content = source_file.read()
+        copy_file.write(content)
```
Bohdan-Salamakha, line 12: unite these two rows

Bohdan-Salamakha (changes requested):
copy_file("cp file.txt file.txt") removed all text from my file but it should do nothing

app/main.py
govorunenko: Done:  
with (open(source_name, "r") as source_file,
         open(copy_name, "w") as copy_file):

app/main.py
govorunenko: Done:

govorunenko: copy_file.write(source_file.read())

Bohdan-Salamakha (approved):

PR #497, opened 2023-07-09
Solution

pnakongit:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_name, copy_file_name = command.split(" ")
```
vasylhnatiuk, line 2: you must check if command.split(" ") has 3 arguments, because code will be broken

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #499, opened 2023-07-11
Develop

Oleksiy-Lyashenko:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_of_arguments_in_command = command.split(" ")
+
+    file_to_copy = list_of_arguments_in_command[1]
+    file_copy = list_of_arguments_in_command[2]
+
+    if file_to_copy != file_copy:
+        with open(file_to_copy, "r") as file_in, \
```
Polyakiv-Andrey, line 8: using brackets instead of

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_of_arguments_in_command = command.split(" ")
```
Polyakiv-Andrey, line 2: split has a space by default, it doesn't need to be specified

Polyakiv-Andrey (changes requested):

PR #500, opened 2023-07-12
Solution

anastmishchuk:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_split = command.split()
+    file_in_name = file_split[1]
+    file_out_name = file_split[2]
```
vsmutok, line 4: Save split result to a separate variable

Validate that the command length is equal exactly to 3 arguments

Validate that command name is exactly "cp"

Use unpacking to access command parts

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_split = command.split()
+    if len(file_split) != 3:
+        print("Invalid command. Usage: cp <source_file> <destination_file>")
+        return
+    _, file_in_name, file_out_name = file_split
```
vasylhnatiuk, line 6:
```suggestion
    command_name, file_in_name, file_out_name = file_split
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+    if file_split[0] != "cp":
```
vasylhnatiuk, line 7:
```suggestion
    if command_name != "cp":
```

vasylhnatiuk (changes requested):

PR #502, opened 2023-07-14
Solution

numerodiciannove:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command.split()[1:]
+    file_one = file_names.split()[0]
+    file_two = file_names.split()[1]
+
+    if file_one != file_two:
+        with open(file_one, "r") as file_in, open(file_two, "w") as file_out:
+            file_out.write(file_in.read)
```
ArsenPidhoretskyi, line 8: 1. First of all validate that command.split() have 3 params
2. Also you can do: cmd, source_path, destination_path = command.split()
3. Check that cmd is really cp
4. Check than files are different

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Now is so good! 👍

PR #503, opened 2023-07-14
solution

vkleshko:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file_read, file_write = command.split()
```
ArsenPidhoretskyi, line 2: Before check that command have 3 params. Because if you do: copy_file("cp") it will raise an error

app/main.py
```diff
@@ -1 +1,8 @@
...
+
+    if file_read == file_write:
```
ArsenPidhoretskyi, line 4: Also check that command is really cp

app/main.py
```diff
@@ -1 +1,8 @@
...
+        return
+    with open(file_read, "r") as file_in, \
+            open(file_write, "w") as file_out:
```
```suggestion
    with (
        open(file_read, "r") as file_in,
        open(file_write, "w") as file_out
    ):
```

ArsenPidhoretskyi (changes requested):

PR #504, opened 2023-07-15
Solution

verbik:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_name = command.split()[1]
```
ArsenPidhoretskyi, line 2: split only once.
Validate that command is cp
Also use uzipping:
```
parts = command.split()
# Here check that len of parts is 3
cmd, source_path, destination_path = parts
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_name, new_file_name = command.split()
```
ArsenPidhoretskyi, line 2: First of all check that command.split() have 3 params

app/main.py
```diff
@@ -1 +1,11 @@
...
+    if command != "cp":
+        return
+    file_content = ""
+    if file_name == new_file_name:
+        return
+    with open(file_name, "r") as f:
+        file_content = f.read()
+    with open(new_file_name, "w") as f:
+        f.write(file_content)
```
ArsenPidhoretskyi, line 11: Use open in one with
```suggestion
    if file_name == new_file_name:
        return
    with (
        open(file_name, "r") as source
        open(new_file_name, "w") as destination
    ):
        destination.write(source.read())
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #506, opened 2023-07-15
solution

vkleshko:


vsmutok (changes requested):
Make sure you check that the cp input command is correct.

PR #508, opened 2023-07-17
Solution

tanya-burdenko:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.split()[1] != command.split()[2]:
+        with open(command.split()[1], "r") as file_in, \
+             open(command.split()[2], "w") as file_out:
```
Grayder0152, line 4: use `()` instead of `\`

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.split()[1] != command.split()[2]:
```
Grayder0152, line 2: check that command has 3 args

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3:
+        original = command.split()[1]
```
ArsenPidhoretskyi, line 3: Also check that command is cp 
And use unzipping

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3 and command.startswith("cp"):
```
ArsenPidhoretskyi, line 2: But for example if command will equal to "cpcpcpcpcpcppcpc file.txt file2.txt"
It will return true. Due to task command must be "cp file.txt file2.txt"

app/main.py
```diff
@@ -1 +1,6 @@
...
+        original, copy = command.split()[1], command.split()[2]
```
ArsenPidhoretskyi, line 3: Use only once split
```suggestion
        original, copy = command.split()[1:]
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3 and command.split()[0] == "cp":
+        original, copy = command.split()[1], command.split()[2]
+        if original != copy:
```
```suggestion
    if len(command.split()) == 3:
        cmd, original, copy = command.split()
        if cmd == "cp" and original != copy:
```

ArsenPidhoretskyi (changes requested):

app/main.py
tanya-burdenko: wow, that's much better, thanks

vasylhnatiuk (approved):

PR #509, opened 2023-07-18
Develop

mwellick:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[1] == command[2]:
+        pass
+    with open(command[1], "r") as file_from:
+        with open(command[2], "w") as file_to:
+            file_to.write(file_from.read())
```
YukihiroSM, line 7: use list unpacking instead of indexes. 
Check, that command is `cp`

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+
+print(copy_file("cp file.txt new_file.txt"))
```
YukihiroSM, line 10: remove this

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[1] == command[2]:
+        pass
```
YukihiroSM, line 4: This will not stop the function execution. This if does nothing

app/new_file.txt
YukihiroSM, line 1: Remove test files from git. Only `py` file should be in git

app/file.txt
YukihiroSM, line 1: Remove test files from git. Only py file should be in git

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_from, file_to = command.split()
```
YukihiroSM, line 2: check, that command is cp

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if file_from == file_to:
+        return
+    with open(file_from, "r") as file_from:
+        with open(file_to, "w") as file_to:
```
YukihiroSM, line 6: you can do it using one `with`
```suggestion
    with (open(file_from, "r") as file_in,
             open(file_to, "w") as file_out):
```

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_from, file_to = command.split()
```
olena-hul, line 2: Check command length before unpacking. It should be equal exactly to 3 parts

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command) != 3:
```
olena-hul, line 2: 1 - split
2 - validate length of splitted command
3 - unpack

olena-hul (changes requested):

app/main.py
mwellick: oops,sorry

EdAlekseiev (approved):
Good job

PR #510, opened 2023-07-20
Solution

AlexIzmashkin:


YukihiroSM (approved):
Good job! Tiny comment: you can create a separate variable like: splitted_command = command.split(). SO you will not need to perform split twice.

PR #511, opened 2023-07-20
solution

vsafian:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, file_name, new_name = command.split(" ")
```

app/main.py
```diff
@@ -1 +1,16 @@
...
+    if (command_name == "cp"
+            and new_name != file_name):
+        with open(f"{file_name}") as file:
+            content = file.read()
+        with open(f"{new_name}", "w") as new_file:
```
vasylhnatiuk, line 7: try to use only one `with`

app/main.py
```diff
@@ -1 +1,16 @@
...
+            new_file.write(content)
+
+
+if __name__ == "__main__":
```
vasylhnatiuk, line 11: why do you need below code?

app/main.py
ArsenPidhoretskyi: Not fixed

app/main.py
ArsenPidhoretskyi: Not fixed

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, file_name, new_name = command.split(" ")
+    if (command_name == "cp"
+            and new_name != file_name):
+        with open(f"{file_name}") as file:
```
ArsenPidhoretskyi, line 5: f-string is redudant

app/main.py
ArsenPidhoretskyi: Also " " is default due to split

Grayder0152 (approved):

PR #513, opened 2023-07-23
Solution

MarinaKhrunina:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import shutil
```
vsmutok, line 1: don't use `shutil`
create your custom solution

app/main.py
```diff
@@ -1 +1,17 @@
...
+    if len(command_parts) != 3:
+        raise ValueError("Invalid command format")
+
+    _, source_file, destination_file = command_parts
+
+    if source_file == destination_file:
+        return
+
+    with open(source_file, "r") as file_in, \
+            open(destination_file, "w") as file_out:
```
vsmutok, line 16: better to use `()` then `\`

vsmutok (changes requested):

app/main.py
MarinaKhrunina: ok

app/main.py
MarinaKhrunina: where exactly? I'd try but in every cases was a mistake

vasylhnatiuk: with (
        open(file_to_copy, "r") as file_in,
    ):

vasylhnatiuk (changes requested):
you pushed nothing

app/main.py
```diff
@@ -1,17 +1,8 @@
-import shutil
-
-
 def copy_file(command: str) -> None:
     command_parts = command.split()
-
-    if len(command_parts) != 3:
```
vasylhnatiuk: why did you delete this?

app/main.py
```diff
@@ -1,17 +1,8 @@
...
-    if source_file == destination_file:
-        return
-
-    with open(source_file, "r") as file_in, \
-            open(destination_file, "w") as file_out:
-        shutil.copyfileobj(file_in, file_out)
+    if command_parts[1] == command_parts[2]:
+        pass
+    if command_parts[0] == "cp":
+        with open(command_parts[1], "r") as file_in,\
```
vasylhnatiuk, line 6: you didn't do it like i wrote you

app/main.py
vasylhnatiuk: return and below code pls

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
you pushed nothing

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    if len(command_parts) != 3:
+        raise ValueError("Invalid command format")
+    if command_parts[1] == command_parts[2]:
+        pass
```
ArsenPidhoretskyi, line 6: What pass will do?

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    if len(command_parts) != 3:
+        raise ValueError("Invalid command format")
+    if command_parts[1] == command_parts[2]:
```
```suggestion
    cmd, source_path, destination_path = command_parts
    if command_parts[1] == command_parts[2]:
```
Use unpacking

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #514, opened 2023-07-28
Solution

jinjo0222988:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split(" ")
```
ArsenPidhoretskyi, line 2: 1. space is default
2. Check that command.split() have exactly 3 parameters

app/main.py
```diff
@@ -1 +1,9 @@
...
+    command = split_command[0]
+    file_out = split_command[1]
+    file_in = split_command[2]
```
ArsenPidhoretskyi, line 5: Use inline unpacking

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
+    command, file_out, file_in = split_command
```
vsmutok, line 3: Check that split_command have exactly 3 parameters

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_out, file_in = command.split()
```
vasylhnatiuk, line 2: Check that split_command have exactly 3 parameters

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: before this line

vasylhnatiuk: command must have 3 arguments, because code will be broken if will be less then 3 here.  check like len(command) == 3 before

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #515, opened 2023-07-28
solution

H1kki:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    command, source_file, destination_file = command_parts
+
+    if (
```
Polyakiv-Andrey, line 5: rewrite this if so that when its condition is met, the file is copied

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #517, opened 2023-07-30
resloved

VladyslavShepilov:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    source, destination = command.split()[1:]
```
ArsenPidhoretskyi, line 2: You need to check that command have 3 parameters
Also check that command is cp

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #519, opened 2023-08-01
'Solution'

struhanets:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    todolist = command.split(" ")
```
ArsenPidhoretskyi, line 2: space is default parameter for split
Also validate that command.split() have 3 parameters

app/main.py
```diff
@@ -1 +1,13 @@
...
+    file_name = todolist[1]
+    file_copy_name = todolist[2]
+    with open(file_name, "w") as some_info:
+        some_info.write(input())
```
ArsenPidhoretskyi, line 6: This is so strange. This code is not needed

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    todolist = command.split(" ")
+    file_name = todolist[1]
+    file_copy_name = todolist[2]
```
ArsenPidhoretskyi, line 4: Use unpacking

app/main.py
```diff
@@ -1 +1,13 @@
...
+    with open(file_name, "w") as some_info:
+        some_info.write(input())
+    if file_name != file_copy_name:
+        with open(file_name, "r") as content_out, \
+                open(file_copy_name, "w") as content_in:
```
```suggestion
        with (
            open(file_name, "r") as content_out,
            open(file_copy_name, "w") as content_in
        ):
```

ArsenPidhoretskyi (changes requested):

app/main.py
struhanets: I can't do it in python 3.8

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if "cp" in command:
```
ArsenPidhoretskyi, line 2: "hello cp.txt file.txt"
will return true. But command is not cp

app/main.py
```diff
@@ -1 +1,13 @@
...
+        todolist = command.split()
+        *other, file_name, file_copy_name = todolist
```
```suggestion
        parts = command.split()
        cmd, file_name, file_copy_name = parts
```

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    todolist = command.split()
```
ArsenPidhoretskyi, line 3: todolist is wrong name
rename to parts

app/main.py
```diff
@@ -1 +1,13 @@
...
+    cmd, file_name, file_copy_name = todolist
+    if cmd == "cp":
+        if file_name != file_copy_name:
```
```suggestion
    if cmd == "cp" and file_name != file_copy_name:
```
Use and instead of two ifs

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1,7 +1,7 @@
 def copy_file(command: str) -> None:
 
-    todolist = command.split()
-    cmd, file_name, file_copy_name = todolist
+    parts = command.split()
+    cmd, file_name, file_copy_name = parts
```
vasylhnatiuk, line 4: if len(parts) less then 3, code will be broke here. add an additional check before

vasylhnatiuk (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #521, opened 2023-08-03
Solution

Anastasia-Su:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted = command.split()
```
vasylhnatiuk, line 2: You must check if splitted has three arguments because if not, the code will fail.

vasylhnatiuk: command_name, filename, copy_filename = command.split() and you can do like this

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, filename, copy_filename = command.split()
```
Grayder0152, line 2: before it check that length of `command.split()` must be 3

Grayder0152 (changes requested):

vasylhnatiuk (approved):

PR #524, opened 2023-08-07
Solution

bogdandovgan:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    file_copy = command[1]
```
vasylhnatiuk, line 3: check if `command `has 3 elements before it

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3:
+        print("Enter the correct command.")
+        return
+    file_copy = command[1]
+    file_paste = command[2]
```
vsmutok, line 7: use unpacking 

`_ file_copy, file_paste = command`

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3:
+        print("Enter the correct command.")
+        return
+    _, file_copy, file_paste = command
```
ArsenPidhoretskyi, line 6: Also validate that command is really cp
Because as for me if the command will be:
"hello file.txt file2.txt"
Is we need a copy file in this case?

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (changes requested):
Changes requested

app/main.py
```suggestion
    cmd, file_copy, file_paste = command
```

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if file_copy != file_paste and _ == "cp":
```
```suggestion
    if file_copy != file_paste and cmd == "cp":
```

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #525, opened 2023-08-08
Loading and dumping into file

IProskurnytskyi:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    separate = command.split()
```
pavlejviki, line 2: You must check if separate has three arguments because if not, the code will fail.

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
...
+    old_file = separate[1]
+    new_file = separate[2]
+    if len(separate) == 3 and separate[0] == "cp" and old_file != new_file:
```
pavlejviki, line 5: The length check should be done before assigning old_file and new_file to avoid potential IndexErrors.

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #526, opened 2023-08-09
Solution

AndriiIshchenko:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_in_list = command.split(" ")
```
```suggestion
    command_in_list = command.split()
```

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if command_in_list[0] == "cp" and command_in_list[1] != command_in_list[2]:
```
vsmutok, line 3: check if command_in_list has 3 elements before it

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_in_list = command.split()
+    if (len(command_in_list) == 3
```
vsmutok, line 3:
```suggestion
    if (len(command.split()) == 3:
            command_name, filename, copy_filename = command.split()
```

vsmutok: it looks better

vsmutok (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #527, opened 2023-08-10
Solution

OlPlaksa:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+
+    if command[1] != command[2]:
```
vsmutok, line 4: also check  if command[1] == cp

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[0] == "cp":
```
vsmutok, line 3: check if command has 3 elements before it

vsmutok: also you can use unpacking 
`command_name, file_name, copy_file_name = command.split()`

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍

PR #529, opened 2023-08-12
'Solution'

usKos:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    args = command.split(" ")
```
YukihiroSM, line 2: Check, that user entered exactly 3 words in command

app/main.py
```diff
@@ -1 +1,9 @@
...
+    com = args[0]
+    file_copy = args[1]
+    new_file = args[2]
```
YukihiroSM, line 5:
```suggestion
    com, file_copy, new_file = args
```

YukihiroSM (changes requested):

YukihiroSM (approved):
Nicely done! Thanks!

PR #531, opened 2023-08-15
Solution

eduardhabryd:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    words = command.split()
+    first_file, second_file = words[1], words[2]
```
Kroll410, line 3: Whast if command doesn't containt spaces?

Kroll410 (changes requested):

app/main.py
eduardhabryd: > Whast if command doesn't containt spaces?

Improved, I also wrote some test for the function, if you have spare time I'd be very thankful if you check and review them

PR #533, opened 2023-08-17
Develop

SuliganLP:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_names = command.split(" ")
+    first_file, second_file = files_names[1], files_names[2]
```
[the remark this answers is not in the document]
SuliganLP: Done!

vasylhnatiuk (changes requested):
image
fix tetss

vasylhnatiuk (changes requested):
image

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_names = command.split()
+    first_file, second_file = files_names[1], files_names[2]
```
vasylhnatiuk, line 3: check if files_names has 3 elements **before** it, because code will be failed if files_names has less then 2 arguments

vasylhnatiuk (changes requested):

app/main.py
SuliganLP: Got it!

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_names = command.split()
```
Grayder0152, line 2: not good name of this split (`cp` - is it file name too ?)

Grayder0152 (changes requested):

app/main.py
SuliganLP: Hi! How about "arguments"? Is this good name?

vsmutok (approved):
Good Job 👍

PR #546, opened 2023-08-22
Develop

WoyDiam:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) <= 4 and parts[0] != "cp":
```
```suggestion
    if len(parts) != 3 and parts[0] != "cp":
```

app/main.py
```diff
@@ -1 +1,21 @@
...
+        print("Invalid command format")
+        return
+
+    s_file = parts[1]
+    d_file = parts[2]
```
ArsenPidhoretskyi, line 8: No need to use abbreviations

ArsenPidhoretskyi: Also it is not a file, it is path

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #565, opened 2023-08-23
Solution

ivanzh13:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+class InvalidCommandError(Exception):
+    ...
+
+
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[0] != "cp":
```
vasylhnatiuk, line 7: check also if len(command_list) == 3

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+class InvalidCommandError(Exception):
+    ...
```
vsmutok, line 2: better to add docstring

vsmutok (approved):

PR #573, opened 2023-08-29
'Solution 29.08.23'

TymchenkoV:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, file, new_file = command.split()
```

vasylhnatiuk (changes requested):

PR #574, opened 2023-08-30
Solutiin

AllaKuksa:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    cd, file_name, new_file = command.split(" ")
```
ArsenPidhoretskyi, line 3: Before better to check that command.split() have exactly 3 parameters)
Also no need to write: command.split(" ") space is default. Just use command.split()

app/main.py
```diff
@@ -1 +1,9 @@
...
+
+    if file_name == new_file:
+        return
+    else:
+        with open(file_name, "r") as file_in, open(new_file, "w") as file_out:
+            file_out.write(file_in.read())
```
```suggestion
    with open(file_name, "r") as file_in, open(new_file, "w") as file_out:
        file_out.write(file_in.read())
```

ArsenPidhoretskyi: else: 
is not needed in this case

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    if len(command) == 3:
```
```suggestion
    if len(command.split()) == 3:
```

app/main.py
```diff
@@ -1 +1,10 @@
...
+        cd, file_name, new_file = command.split()
+
+        if file_name == new_file:
```
ArsenPidhoretskyi, line 6: also check that cd is really "cp"

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    if len(command.split()) == 3 and command.split()[0] == "cp":
+        cd, file_name, new_file = command.split()
```
```suggestion
    parts = command.split()
    if len(parts) == 3 and parts[0] == "cp":
        cd, file_name, new_file = parts
```
Better to do splitting one time instead of 3

ArsenPidhoretskyi (changes requested):
Require small fix

ArsenPidhoretskyi (approved):
Well done!

PR #577, opened 2023-09-03
Solution

sachavskyi:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(filename: str) -> None:
+    file_for_copy = filename.split(" ")[1]
+    new_file = filename.split(" ")[2]
```
Bohdan-Salamakha, line 3: don't call split method twice
create a separate variable and use it on demand

Bohdan-Salamakha (changes requested):
u don't need to recreate a file
copy_file("cp file.txt file.txt")  # Does nothing

viktoria-rybenchuk (changes requested):
you need to go through the checklist and you didn't fix comments from Bohdan

Polyakiv-Andrey (approved):
Good job!

PR #579, opened 2023-09-05
Develop

ladno338:


ArsenPidhoretskyi (changes requested):

vasylhnatiuk (changes requested):
do not forgot to validate that command.split() have 3 parameters and command is cp

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split(" ", 3)
```
```suggestion
    commands = command.split()
```

vasylhnatiuk: it automatically separate to 3 arguments by " "

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #580, opened 2023-09-06
copy file

StepanShevchuk:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_list = command.split(" ")
+    if files_list[1] != files_list[2]:
```
Grayder0152, line 3: before it pls check that command len == 3

Grayder0152 (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_list = command.split(" ")
+    if len(files_list) == 3:
+        if files_list[1] != files_list[2]:
```
ArsenPidhoretskyi, line 4: Also you must check that command is really cp

ArsenPidhoretskyi, line 4: And use unpacking instead of indexes:
cmd, source_path, destination_path = files_list

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_list = command.split(" ")
```
ArsenPidhoretskyi, line 2: Rename files_list to command_parts

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #582, opened 2023-09-08
task completed (1)

DaniilHinzburh:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
+
+    if command[1] == command[2]:
```
vasylhnatiuk, line 4: you should check if `command ` has 3 atrguments before this line

vasylhnatiuk: and check if command name is `cp`

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,8 +1,23 @@
...
+
+
 def copy_file(command: str) -> None:
+    if command[:2] != "cp":
+        raise NameCommandError
     command = command.split(" ")
 
-    if command[1] == command[2]:
-        pass
+    if len(command) != 3:
```
vasylhnatiuk, line 18: you must raise CommandError or end function if len(command) != 3:

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,27 @@
...
+
+    if len(command) != 3:
+        raise CommandError
+
+    if command[1] == command[2]:
+        pass
+
+    with open(command[1], "r") as data_file:
+        data = data_file.read()
+        with open(command[2], "w") as new_file:
```
vasylhnatiuk, line 26: use only one with

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,26 @@
...
+
+
+class CommandError(CustomError):
+    """The error of missing three elements in the team"""
+
+
+def copy_file(command: str) -> None:
+    if command[:2] != "cp":
+        raise NameCommandError
+    command = command.split(" ")
```
vsmutok, line 16:
```suggestion
    command = command.split()
```

app/main.py
```diff
@@ -1 +1,26 @@
...
+class NameCommandError(CustomError):
+    """Command name is not cp"""
+
+
+class CommandError(CustomError):
+    """The error of missing three elements in the team"""
+
+
+def copy_file(command: str) -> None:
+    if command[:2] != "cp":
```
vsmutok, line 14: why `command[:2] != "cp"` ????
what is meant here?

vsmutok (changes requested):

app/main.py
vsmutok: the idea is good, but to check whether the command starts with `cp` you need to write like this `command[0] == "cp"`

vsmutok (changes requested):

app/main.py
vsmutok: not fixed

vsmutok (changes requested):

vsmutok (approved):
Great 👍

PR #587, opened 2023-09-11
py-copy-file

stamaksim:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> str:
+    old_file, new_file = command.split()[1:]
```
ArsenPidhoretskyi, line 2: Also you must validate that command really have 3 parameters and command is "cp"

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (changes requested):
fix tests

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if len(old_file) != 3 and old_file != "cp":
```
```suggestion
    if len(command) != 3 and old_file != "cp":
```

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> str:
+    old_file, new_file = command.split()[1:]
```
vasylhnatiuk, line 2: and this must after 3 line

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> str:
+    if len(command) != 3 and command != "cp":
```
```suggestion
    if len(command) != 3 and command[0] != "cp":
```

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: cp is the first argument in command

vasylhnatiuk (approved):

PR #588, opened 2023-09-14
Solution

endernaga:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(comand: str) -> None:
+    file1, file2 = comand.split()[1], comand.split()[2]
```
EdAlekseiev, line 2: You can use unpacking instead of getting elements by index

```suggestion
    cmd, file1, file2 = comand.split()
```

app/main.py
```diff
@@ -1 +1,5 @@
...
+    if not file1 == file2:
```
EdAlekseiev, line 3: Also, check that `comand` starts with `cp`

EdAlekseiev (changes requested):

Bohdan-Salamakha (approved):

PR #589, opened 2023-09-15
Solution-py-copy-file

Nadiia-developer:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split(" ")
+    print(parts)
```
Dimosphen1, line 3: Remove this `print` from your code, it is not needed.

app/main.py
```diff
@@ -1 +1,17 @@
...
+    source_path = parts[1]
+    destination_path = parts[2]
+
+    if source_path == destination_path:
+        return
+
+    with open(source_path, "r") as file_in, \
+            open(destination_path, "w") as file_out:
+        for line in file_in:
+            file_out.write(line)
```
Dimosphen1, line 13: You do not need to make a `for` loop here, use the `.writelines` method instead.

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split(" ")
+    print(parts)
+    source_path = parts[1]
```
Dimosphen1, line 4: Make sure that the first part of the command is `cp` in the body of our function.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

app/main.py
```diff
@@ -1 +1,19 @@
...
+        return
+
+    source_path = parts[1]
+    destination_path = parts[2]
+
+    if source_path == destination_path:
+        return
+
+    with open(source_path, "r") as file_in, \
+         open(destination_path, "w") as file_out:
```
viktoria-rybenchuk, line 14: try to avoid using backslash  when you move to new line

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split(" ")
+
+    if len(parts) < 3 or parts[0] != "cp":
```
viktoria-rybenchuk, line 4: what if `len(parts) > 3`? Does it work correctly?

viktoria-rybenchuk (changes requested):

app/main.py
Nadiia-developer: Check if the first part of the command is "cp"

viktoria-rybenchuk (approved):
GJ!

PR #593, opened 2023-09-18
'Solution'

o-sapohova:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
```
olena-hul, line 2: Validate command length

Use unpacking to access command parts (a, b = [1,2])

olena-hul (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
...
+    origin_file, new_file = command[1], command[2]
+    if origin_file != new_file:
```
EdAlekseiev, line 4: Also, check that the command starts with "cp"

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    start_command, origin_file, new_file = command[0], command[1], command[2]
```
EdAlekseiev, line 3: Use unpacking here instead of getting elements by index

EdAlekseiev, line 3:
```suggestion
    start_command, origin_file, new_file = command
```

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
```
olena-hul, line 2: There is no validation for the command

Command name should be exactly "cp" and command should consist of 3 parts

olena-hul (changes requested):

Bohdan-Salamakha (approved):

PR #595, opened 2023-09-20
solution

MykolaKrasulin:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split()
+
+    with (
```
Dimosphen1, line 4: Make sure that the command includes `cp` in the right place and that there are exactly two filenames that are passed in a command as mentioned [in the checklist](https://github.com/MykolaKrasulin/py-copy-file/blob/develop/checklist.md#code-efficiency).

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split()
+    if command_split[0] == "cp" and len(command_split) == 3:
```
viktoria-rybenchuk, line 3: first check for this `It must do nothing in case the user is trying to copy file to file with the same name`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split()
+    if command_split[1] != command_split[2]:
```
viktoria-rybenchuk, line 3: you can use unpacking here but first check if len(command_split) == 3
After that you need check if command is `cp` and file1 != file2 and than copy file if not do nothing

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split()
```
pavlejviki, line 2: better to use unpacking and save the parts into variables with  meaningful names. command_split[0], command_split[1] are not informative

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_names = command.split()
+    if len(files_names) == 3:
+        if files_names[0] == "cp" and files_names[1] != files_names[2]:
+            with (
+                open(files_names[1], "r") as file_in,
+                open(files_names[2], "w") as file_out
+            ):
+                if files_names[1] != files_names[2]:
```
viktoria-rybenchuk, line 9: this is redundant, since this condition is already checked in the outer if statement, it doesn't need to be checked again

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Good job!

PR #596, opened 2023-09-20
solution

antonnech2309:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split(" ")
```
viktoria-rybenchuk, line 2: Make sure you check that the `cp` input command is correct.

viktoria-rybenchuk (changes requested):

app/main.py
pavlejviki, line 2: better to use unpacking and save the parts into variables with meaningful names. files[1] , files[2] are not informative

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #597, opened 2023-09-21
Solution

MishkaGAMII:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, source, new = command.split(" ")
```
ArsenPidhoretskyi, line 2: Also before this validate that command.split() have 3 parameters
if len(command.split()) != 3:
    return
....

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split(" ")) != 3:
```
```suggestion
    if len(command) != 3:
```

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
do it

ArsenPidhoretskyi (approved):
Well done!

PR #599, opened 2023-09-22
implemented py-copy-file task

HlovaOleh:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    content = command.split()
+    if content[1] != content[2]:
```
EdAlekseiev, line 6: Check that the command starts with "cp"

app/main.py
```diff
@@ -1 +1,9 @@
...
+        with open(content[1], "rb") as src_file, \
+                open(content[2], "wb") as dest_file:
```
EdAlekseiev, line 8: Use `()` instead of `\`

app/main.py
```diff
@@ -1 +1,9 @@
...
+            shutil.copyfileobj(src_file, dest_file)
```
EdAlekseiev, line 9: Try to solve this task without additional library

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    content = command.split()
+    if content[0] == "cp" and content[1] != content[2]:
```
YukihiroSM, line 3: If command length is less than 3 - you will get an exception

YukihiroSM: Also, it's better to use unpacking here instead of using indexes.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
...
+    action, source_file, destination_file = command.split()
+    if action == "cp" and source_file != destination_file:
+        with open(source_file, "rb") as src_file, (
+                open(destination_file, "wb")
+        ) as dest_file:
+            while True:
+                data = src_file.read()
+                if not data:
+                    break
+                dest_file.write(data)
```
Bohdan-Salamakha, line 11: u don't need the 'while' cycle here

Bohdan-Salamakha (changes requested):

Bohdan-Salamakha (approved):

PR #601, opened 2023-09-25
solution

EvgenMilokumov:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    find_txt = command.index("txt")
```
ArsenPidhoretskyi, line 2: why you need txt search?

ArsenPidhoretskyi (changes requested):
Wrong realizations. If file name will be:
txt.txt code will not work

app/main.py
EvgenMilokumov: deleted

EvgenMilokumov: realization has been changed, thanks

app/main.py
```diff
@@ -1,10 +1,7 @@
...
-    first_file_name = command[3 : (find_txt + 3)]
-    second_file_name = command[(find_txt + 4) :]
-
-    with open(first_file_name, "r") as file_in, open(
-        second_file_name, "w"
-    ) as file_out:
-        file_in_content = file_in.read()
-        file_out.write(str(file_in_content))
+    splited_command = command.split()
+    if splited_command[1] != splited_command[2]:
```
vasylhnatiuk, line 3: what if `splited_command == []`?

vasylhnatiuk (changes requested):

app/main.py
EvgenMilokumov: Function take only one argument command, **that is string with command cp, file name to copy and new file name, separated by spaces.**

I thought this parameter isn't empty. 

Please specify, do you mean that I need to add a check like (if splited_command != []?) or my realization isn't correct at all?

EvgenMilokumov: added checks

vasylhnatiuk (approved):

vasylhnatiuk (approved):

PR #602, opened 2023-09-26
Solution

Filllick:


ArsenPidhoretskyi (changes requested):

Filllick:
First checks if command starts with "cp" and has 3 parameters, then copies all lines from basic file to a new one.

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_data = command.split()
+    if command_data[0] != "cp" or len(command_data) != 3:
+        return
+
+    with open(command_data[1], "r") as file_in, \
```
vasylhnatiuk, line 6: use () instead of \

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #603, opened 2023-09-26
Solution

valikdeb:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    source_file_name = command.split()[1]
+    new_file_name = command.split()[2]
```
viktoria-rybenchuk, line 3: It's better to use unpacking to improve readability and avoid multiple splits

viktoria-rybenchuk (changes requested):
adjust your code in relation to the checklist

viktoria-rybenchuk (changes requested):
Make sure you check that the cp input command is correct and band check that the length of the 'command' is 3.

viktoria-rybenchuk (changes requested):
You didn't fix it. Where do you check command?

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+class NameCommandError(Exception):
+    """Command name is not cp"""
+
+
+class CommandError(Exception):
+    """The error of missing three elements in the team"""
```
viktoria-rybenchuk, line 6: In which team? Write descriptive error message

app/main.py
```diff
@@ -1 +1,19 @@
...
+
+
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
```
viktoria-rybenchuk, line 10: You can store the result of command.split() in a variable and then use it.

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,21 @@
...
+
+    if len(command_split) != 3:
+        raise CommandError
+
+    name_command, source_file_name, new_file_name = command_split
+
+    if name_command != "cp":
+        raise NameCommandError
+
+    with open(source_file_name) as file, open(new_file_name, "w") as new_file:
```
pavlejviki, line 20: It must do nothing in case the user is trying to copy file to file with the same name.

pavlejviki (changes requested):

PR #604, opened 2023-09-27
Develop

tiron-vadym:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from shutil import copyfileobj
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
```
Dimosphen1, line 6: Check that the first part is `"cp"` to ensure that the input is correct.

app/main.py
```diff
@@ -1 +1,15 @@
...
+        return
+
+    cp, source_file, destination_file = parts
+    if source_file == destination_file:
```
Dimosphen1, line 10: Write one if statement that will abandon the execution, there is no need to split logic in two or more if conditions here.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+from shutil import copyfileobj
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
```
viktoria-rybenchuk, line 5: Make sure you check that the cp input command is correct.

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if len(parts) == 3 and parts[1] != parts[2]:
```
viktoria-rybenchuk, line 6: try to use unpacking for better readability

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+from shutil import copyfileobj
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+    source_filename, destination_filename = parts[1], parts[2]
```
Dimosphen1, line 6: You do not need to perform unpacking with the use of indexes, write all results to a separate variable at once. Besides that, you will use the first element as well later in the code.
```suggestion
    command, source_filename, destination_filename = parts
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if (len(parts) == 3 and parts[0] == "cp"
+            and source_filename != destination_filename):
```
Dimosphen1, line 8: You could replace the use of two `and` with `all([...])`, where you can pass three conditions in a list.

Dimosphen1 (changes requested):

PR #607, opened 2023-10-01
Solution

SerhiiBudko:


app/main.py
```diff
@@ -1 +1,10 @@
...
+def copy_file(command: str) -> None:
+    file_name = command.split()
+    if len(file_name) == 3:
+        if file_name[0] == "cp" and file_name[1] != file_name[2]:
+            with (
+                open(file_name, "r") as file_in,
+                open(file_name, "w") as file_out
+            ):
+                file_to_read = file_in.read()
+                file_out.write(file_to_read)
```
Dimosphen1, line 10: Do not create a separate variable `file_to_read`, call `file_in.read()` inside of `write()` directly, you won't need to allocate space for this variable then.

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_name = command.split()
+    if len(file_name) == 3:
```
Dimosphen1, line 3: Write one if condition and avoid nesting in your code when possible. You could use `all([...])` and place all your conditions inside of the list to organize your code better.

app/main.py
```diff
@@ -1 +1,10 @@
...
+        if file_name[0] == "cp" and file_name[1] != file_name[2]:
```
Dimosphen1, line 4: You could perform unpacking at the beginning of your code after the `split()` operation so that you would avoid indexing. It could make code more descriptive.

app/main.py
```diff
@@ -1 +1,10 @@
...
+            with (
+                open(file_name, "r") as file_in,
```
Dimosphen1, line 6: You are passing the list `file_name` as a first argument to `open`, pass the string with the required filename instead.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1,10 +1,23 @@
...
+    try:
+        with (
+            open(main_file, "r") as file_in,
+            open(destination_file, "w") as file_out
+        ):
+            file_out.write(file_in.read())
+        return "Files were copied successfully!"
+    except FileNotFoundError as e:
+        print(f"Error: {e}")
+    except Exception as e:
```
pavlejviki, line 22: Using a broad Exception can catch any error that arises, which is generally not recommended

app/main.py
```diff
@@ -1,10 +1,23 @@
...
+        return "Files are the same!"
+
+    try:
+        with (
+            open(main_file, "r") as file_in,
+            open(destination_file, "w") as file_out
+        ):
+            file_out.write(file_in.read())
+        return "Files were copied successfully!"
+    except FileNotFoundError as e:
```
pavlejviki, line 20: it's entirely reasonable to let the exception raise  and potentially terminate the program. no need to catch it and just print the message

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> str:
+    try:
+        parts = command.split()
+
+        if len(parts) != 3:
+            raise ValueError("Invalid command format")
+
+        _, source_file, destination_file = parts
```
Dimosphen1, line 8: Ensure that the first part of the command is `cp`, do not skip its value.

app/main.py
Dimosphen1: There is no need to catch an exception and raise it further. Remove try-except block from your code.

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> str:
```
Dimosphen1, line 1: `None` should be returned as a result of the script, not `str`.

Dimosphen1 (changes requested):

pavlejviki (approved):
Well done!

PR #608, opened 2023-10-02
Solution

asdadaversa:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    command = command.split()
```
pavlejviki, line 3: add additional checks to make sure that the  command  consists of three parts and that it starts with "cp"

app/main.py
```diff
@@ -1 +1,5 @@
...
+    with open(command[1], "r") as file_in, open(command[2], "w") as file_out:
```
pavlejviki, line 4: It must do nothing in case the user is trying to copy file to file with the same name.

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    command = command.split()
+    if command[1] == command[2]:
```
pavlejviki, line 4: you should check the length of commands before indexing

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #609, opened 2023-10-03
Solution

Lebwa1337:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_name_number1 = command.split(" ")[1]
+    file_name_number2 = command.split(" ")[2]
```
viktoria-rybenchuk, line 3: you can use unpacking

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
```
viktoria-rybenchuk, line 1: where do you check  that the `cp ` input command is correct.? Read checklist

app/main.py
```diff
@@ -1 +1,7 @@
...
+    file_name_number1 = command.split(" ")[1]
+    file_name_number2 = command.split(" ")[2]
+    if not file_name_number1 == file_name_number2:
```
viktoria-rybenchuk, line 4:
```suggestion
    if  file_name_number1 != file_name_number2:
```

viktoria-rybenchuk (changes requested):

app/main.py
Lebwa1337: Just like if  "cp" in command? or something else?

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    comm, file_name_number1, file_name_number2 = command.split()
```
pavlejviki, line 2:
```suggestion
    command, file_1, file_2 = command.split()
```

app/main.py
pavlejviki: as well you need to check if command consists of three parts

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    comm, file_1, file_2 = command.split()
```
viktoria-rybenchuk, line 2: before unpacking you should check if command consists three parts

viktoria-rybenchuk, line 2: write descriptive variable names as command,  source_file, destination_file

viktoria-rybenchuk (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #610, opened 2023-10-04
Implemented copy_file function

sberdianskyi:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
```
ArsenPidhoretskyi, line 2: First of all validate that command.split() have 3 params

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if command_list[1] != command_list[2]:
```
ArsenPidhoretskyi, line 3: Also check that command is cp

app/main.py
```diff
@@ -1 +1,7 @@
...
+        with (open(command_list[1], "r") as file_in,
+              open(command_list[2], "w") as file_out):
+            for line in file_in:
+                file_out.write(line)
```
ArsenPidhoretskyi, line 7: No need this loop

ArsenPidhoretskyi (changes requested):

vasylhnatiuk (approved):

PR #611, opened 2023-10-05
solution

Depaton:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[1] != command[2]:
```
viktoria-rybenchuk, line 3: Use unpacking

viktoria-rybenchuk (changes requested):
Make sure you check that the cp input command is correct.

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3 and command.split()[0] == "cp":
```
pavlejviki, line 2:
```suggestion
    command = command.split()
    if len(command) == 3:
```

app/main.py
```diff
@@ -1 +1,7 @@
...
+        action, file, new_file = command.split()
```
pavlejviki, line 3:
```suggestion
        action, file, new_file = command
```

app/main.py
```diff
@@ -1 +1,7 @@
...
+        if file != new_file:
```
pavlejviki, line 4:
```suggestion
        if action == "cp" and file != new_file:
```

pavlejviki (changes requested):

pavlejviki (approved):
Well done!

PR #615, opened 2023-10-12
Done

DimaPr7:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    time_result = command.split(" ")
+    with open(time_result[1], "r") as current_file,\
```
vasylhnatiuk, line 3: use () instead of \

vasylhnatiuk (changes requested):

app/main.py
DimaPr7: How?

vasylhnatiuk (changes requested):
you must check if command is "cp" and comand has 3 arguments

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    time_result = command.split(" ")
+    if time_result[0] == "cp" and len(time_result) == 3:
+        with open(time_result[1], "r") as current_file,\
```
vasylhnatiuk, line 4:
```suggestion
        with (open(time_result[1], "r") as current_file,
            open(time_result[2], "w+") as new):
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    time_result = command.split(" ")
```
```suggestion
   splitted_command = command.split(" ")
```

vasylhnatiuk: Why time_result? Use more accurate name

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #616, opened 2023-10-13
solution

Stainback:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_items = command.split()
+
+    if command_items[1] != command_items[2]:
```
YukihiroSM, line 4: what if command length is 2?
Also, check that command starts with cp

YukihiroSM (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #617, opened 2023-10-14
'Solution'

TROTPOL:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    vul = command.split()
+    if vul[1] == vul[2] and vul[0] == "cp":
+        return
+    with open(vul[1], "r") as file_in, open(vul[2], "w") as file_out:
+        file_out.write(file_in.read())
```
vsmutok, line 6: Make sure you check that the cp input command is correct.

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    vul = command.split()
```
vsmutok, line 2: `vul` - bad variable  name

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[0] == "cp" and command_list[1] != command_list[2]:
+        with open(command_list[1], "r") as file_in,\
+             open(command_list[2], "w") as file_out:
```
vsmutok, line 5: better use `()` then `\`

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[0] == "cp" and command_list[1] != command_list[2]:
```
vsmutok, line 3: first you need to check if the length of `command_list ` is three

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[0] == "cp" and len(command_list) == 3:
```
vsmutok, line 3: what if we pass empty command?

vsmutok (changes requested):

app/main.py
TROTPOL: The condition does not mention an empty command

vsmutok (approved):
Great 👍

PR #618, opened 2023-10-14
Solution

YuliaHladyshkevych:


prochigor (approved):
It isn't written in the readme file but what is the first file does not exist

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```

LiudmylaKulinchenko (approved):
Good job! 🍬

PR #619, opened 2023-10-15
Solution

prochigor:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, file, file_to_copy = command.split()
+    if command_name != "cd":
+        print("This command not exist")
```
petrykivd, line 4: cp maybe?

petrykivd (approved):

app/main.py
prochigor: oh yes, its a mistake

app/main.py
```diff
@@ -1 +1,14 @@
...
+        return
+    try:
+        with (open(file, "r") as current_file,
+              open(file_to_copy, "x") as new_file):
+            new_file.write(current_file.read())
+    except FileNotFoundError:
+        print(f"There is not file with name {file}")
+        return
+    except FileExistsError:
+        print("This file already exist")
```
LiudmylaKulinchenko, line 14: Are you sure your custom messages in console are more descriptive than default ones?

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, file, file_to_copy = command.split()
+    if command_name != "cp":
+        print("This command not exist")
+        return
```
LiudmylaKulinchenko, line 5: You can use opposite condition here

app/main.py
```diff
@@ -1 +1,14 @@
...
+    try:
+        with (open(file, "r") as current_file,
```
LiudmylaKulinchenko, line 7: You seem to have forgotten to check for this condition:
`It must do nothing if the user tries to copy the file to file with the same.`

LiudmylaKulinchenko (changes requested):

app/main.py
prochigor: I don't, I just use mode x  — writes data to a file and returns an error if the file already exists (exclusive creation); So it's can't copy file with the same name

app/main.py
prochigor: Like that?

app/main.py
prochigor: I thought so

app/main.py
```diff
@@ -1 +1,10 @@
...
+def copy_file(command: str) -> None:
+    command_name, file, file_to_copy = command.split()
+    if command_name == "cp" and file != file_to_copy:
+        try:
+            with (open(file, "r") as current_file,
+                  open(file_to_copy, "w") as new_file):
+                new_file.write(current_file.read())
+        except FileNotFoundError as found_error:
+            print(found_error)
+            return
```
LiudmylaKulinchenko, line 10: You catching all you need in the if condition in 3 line, you can skip try/except

LiudmylaKulinchenko (changes requested):

app/main.py
prochigor: I remove it

Katherine-Greg (approved):

LiudmylaKulinchenko (approved):
Good job!

PR #621, opened 2023-10-15
Solution

IlyaGrynyshyn:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, input_file, output_file = command.split(" ")
```
pavlejviki, line 2: cp is not a descriptive variable name

pavlejviki, line 2: make sure that the command consists of three parts

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
+    if len(command) == 3:
+        function, input_file, output_file = command
+        if function != function or input_file == output_file:
```
viktoria-rybenchuk, line 5: what do you check by this `if function != function`. You should check if command is `cp`

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
+    if len(command) == 3:
+        function, input_file, output_file = command
```
viktoria-rybenchuk, line 4: this is not a `function` it is command. Write correct variable names)

viktoria-rybenchuk (changes requested):

pavlejviki (approved):
Good job!

PR #622, opened 2023-10-15
Solution

Daniil-Pankieiev:


LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, old_file, new_file = command.split()
+    if old_file == new_file or command != "cp":
+        return
```
LiudmylaKulinchenko, line 4: try not to use blank return here, make a backward condition

LiudmylaKulinchenko (changes requested):

app/main.py
Daniil-Pankieiev: Like this?

LiudmylaKulinchenko (approved):
Great! 🚀

PR #623, opened 2023-10-16
Solution

Naz-iv:


mykolamateichuk (approved):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split()
```

app/main.py
```diff
@@ -1 +1,7 @@
...
+
+    if commands[0] == "cp" and commands[1] != commands[2]:
+        with (open(commands[1], "r") as file_in,
+              open(commands[2], "w") as file_out):
+            file_out.writelines(file_in.readlines())
```
LiudmylaKulinchenko, line 7: you can use `write` and `read` here

LiudmylaKulinchenko (changes requested):

app/main.py
Naz-iv: thanks, corrected

LiudmylaKulinchenko (approved):
Great! 🚀

PR #624, opened 2023-10-16
Solution

mykolamateichuk:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_name_to_copy = command.split(" ")[1]
+    new_file_name = command.split(" ")[2]
```
Familenko, line 3: You can use unpacking in one line of code

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+    if file_name_to_copy == new_file_name:
+        return
```
Familenko, line 6: Mayeby better to use != conditions to avoid return. And also what if command will be not "cp"?

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+    with open(file_name_to_copy) as file_to_copy,\
+         open(new_file_name, "w") as new_file:
```
Familenko, line 9: This row will work without "\", becouse it is statment for "with" till the ":"

app/main.py
```diff
@@ -1 +1,10 @@
...
+        new_file.writelines(file_to_copy.readlines())
```
Familenko, line 10: You do not need to unpack data if you are not planning to read or use it. You may directly copy it

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_name_to_copy = command.split(" ")[1]
```
Familenko, line 2: By default split have " "

Familenko (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file_name_to_copy, new_file_name = command.split()
```
LiudmylaKulinchenko, line 2: use more descriptive name for first variable

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Good job! 😸

PR #629, opened 2023-10-16
py-copy-file Solution

RVChornyy:


katyakoziar (approved):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_tuple = tuple(command.split(" "))
```
fsocie7y, line 2: no need to use space inside split method, it uses space by default

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if ((len(command_tuple) != 3
+        or command_tuple[1] == command_tuple[2])
+            or command_tuple[0] != "cp"):
+        return
```
fsocie7y, line 6: try not to use blank return here, make a backward condition

fsocie7y (changes requested):

app/main.py
RVChornyy: without space inside split method "ValueError: empty separator" is beeing rised

fsocie7y (approved):
well done ^_^

PR #630, opened 2023-10-16
Solution py_copy_file

wataschi:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(files_name: str) -> None:
+    command, name_old_file, name_new_file = files_name.split(" ")
+    print(command, name_old_file, name_new_file)
```
viktoria-rybenchuk, line 3: remove this print

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(files_name: str) -> None:
+    command, name_old_file, name_new_file = files_name.split(" ")
```
viktoria-rybenchuk, line 2: First, you need to check the length of file names before unpacking

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(files_name: str) -> None:
+    command_line = files_name.split(" ")
```
viktoria-rybenchuk, line 2: You can simply call the `split()` method without any arguments. By default, it will split the string on whitespace

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (approved):
Great job!

PR #631, opened 2023-10-16
Copy File solution

Lyutillis:


Shatrovskyi (approved):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, initial_file, new_file = command.split()
+    if initial_file == new_file:
+        return
```
fsocie7y, line 4: try not to use blank return here, make a backward condition

fsocie7y (changes requested):
all checks should be passed)

LiudmylaKulinchenko (approved):
Well done!

PR #634, opened 2023-10-17
Solution

katyakoziar:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split("")
+    command_name, source_file, copied_file = command_list
+
+    if len(command_list) != 3:
+        return
+
+    if command_name != "cp":
+        return
```
fsocie7y, line 9: try not to use blank return here, make a backward condition

fsocie7y (changes requested):

fsocie7y (approved):

PR #635, opened 2023-10-17
Solution

Shatrovskyi:


vshvanska (approved):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    commands_data = command.split()
+    first_file = commands_data[1]
+    second_file = commands_data[2]
+
+    if first_file == second_file:
```
fsocie7y, line 7: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    commands_data = command.split()
+    first_file = commands_data[1]
+    second_file = commands_data[2]
```
fsocie7y, line 5: you can unpack these variables in the easiest way, also use the more descriptive names

fsocie7y (changes requested):

PR #636, opened 2023-10-17
Develop

MiskoRuslan:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
```
fsocie7y, line 2: no need to use space inside split method, it uses space by default

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if command_list[1] != command_list[2] and command_list[0] == "cp":
```
fsocie7y, line 3: you can unpack these variables in the easiest way, also use the more descriptive names

fsocie7y (changes requested):

app/main.py
MiskoRuslan: Done :)

fsocie7y (approved):
well done ^_^

PR #637, opened 2023-10-17
Solution

DenPrislipskyi:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    command_type, source_file, copied_file = command_parts
+
+    if len(command_parts) != 3 or command_type != "cp":
+        return
```
fsocie7y, line 6: try not to use blank return here, make a backward condition

fsocie7y (approved):
well done ^_^

PR #639, opened 2023-10-17
Solution

skoval7:


app/main.py
```diff
@@ -1 +1,24 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        print("Invalid command format. Use 'cp source_file destination_file'")
+        return
```
RabenkoYevhenii, line 5: try to use opposite condition to avoid empty return

app/main.py
```diff
@@ -1 +1,24 @@
...
+
+    source_file = parts[1]
+    destination_file = parts[2]
+
+    if source_file == destination_file:
+        print("Source and destination file names are the same."
+              "Nothing to copy.")
+        return
```
RabenkoYevhenii, line 13: Also here

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```

app/main.py
```diff
@@ -1 +1,22 @@
...
+    if len(parts) == 3 and parts[0] == "cp":
+        source_file = parts[1]
+        destination_file = parts[2]
+
+        if source_file == destination_file:
+            print("Source and destination file names are the same."
+                  " Nothing to copy.")
```
LiudmylaKulinchenko, line 9: use opposite condition and do nothing if they are the same `It must do nothing in case the user is trying to copy file to file with the same name.`

app/main.py
```diff
@@ -1 +1,22 @@
...
+            try:
+                with (open(source_file, "rb") as source,
+                      open(destination_file, "wb") as destination):
+                    destination.write(source.read())
+                print(f"File '{source_file}'"
+                      f" copied to '{destination_file}' successfully.")
+            except FileNotFoundError:
+                print(f"File '{source_file}' not found.")
+            except Exception as e:
+                print(f"An error occurred: {e}")
```
LiudmylaKulinchenko, line 20: No need to catch those, your func should just copy file, not handle every error

app/main.py
```diff
@@ -1 +1,22 @@
...
+        else:
+            try:
+                with (open(source_file, "rb") as source,
+                      open(destination_file, "wb") as destination):
+                    destination.write(source.read())
+                print(f"File '{source_file}'"
+                      f" copied to '{destination_file}' successfully.")
+            except FileNotFoundError:
+                print(f"File '{source_file}' not found.")
+            except Exception as e:
```
LiudmylaKulinchenko, line 19: It is a bad practice to catch Exception. It could be even more clear for user if you skip it here

LiudmylaKulinchenko (changes requested):

PR #640, opened 2023-10-17
Solution

RabenkoYevhenii:


fedorov-s-od (approved):
Generally, everything is OK. But in my opinion, I would place the unpacking of "parts" values outside the first "if"-construction. And there would be no need to create a second condition.

PR #642, opened 2023-10-17
Solution

Wonsky1:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        return
+
+    _, filename, new_filename = parts
+    if filename == new_filename:
+        return
```
fsocie7y, line 8: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+    try:
+        with open(filename, "rb") as source,\
```
fsocie7y, line 11: Try to avoid using linebreakers in your code

fsocie7y (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
LiudmylaKulinchenko, line 2: unpack your variables here and then make checks

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if len(parts) == 3:
+        _, filename, new_filename = parts
```
LiudmylaKulinchenko, line 4: what if command is not equal to "cp"? check that

app/main.py
```diff
@@ -1 +1,13 @@
...
+        if filename != new_filename:
+            try:
+                with (open(filename, "rb") as source,
+                      open(new_filename, "wb") as destination):
+                    destination.write(source.read())
+            except FileNotFoundError:
+                print(f"File not found: {filename}")
+            except Exception as e:
+                print(f"An error occurred: {e}")
```
LiudmylaKulinchenko, line 13: You don't need to catch errors here. You will just have a regular error with regular traceback, it's really fine. You should handle that only if you want to change logic somehow, no need to just rewrite messages

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if len(parts) == 3:
+        _, filename, new_filename = parts
+        if filename != new_filename:
+            try:
+                with (open(filename, "rb") as source,
+                      open(new_filename, "wb") as destination):
+                    destination.write(source.read())
+            except FileNotFoundError:
+                print(f"File not found: {filename}")
+            except Exception as e:
```
LiudmylaKulinchenko, line 12: It's a bad practice to catch Exception

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
well done ^_^

PR #643, opened 2023-10-17
Solution

roffi37:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    act, first_file, second_file = command.split()
+    if act == "cp":
+        if first_file[1] != second_file[2]:
```
prochigor, line 4: you can write these conditions in one lone with "and"

prochigor (approved):

app/main.py
LiudmylaKulinchenko, line 4: why do you slice file names? 🤔 Just check if first_file != second_file

LiudmylaKulinchenko (changes requested):

fsocie7y (approved):
GJ 🚀

PR #651, opened 2023-10-23
Solution

kravetb:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_command = command.split()
+    if list_command[1] == list_command[2]:
```
EdAlekseiev, line 3: Also, check that command starts with cp

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_command = command.split()
+    if list_command[1] == list_command[2] and list_command[0] == "cp":
+        return
+    with (open(list_command[1], "r") as file_in,
+          open(list_command[2], "w") as file_out):
```
EdAlekseiev, line 6: You can unpack list_command instead of getting value by index. It makes your code clearer.

EdAlekseiev (changes requested):
Good job

app/main.py
kravetb: > You can unpack list_command instead of getting value by index. It makes your code clearer.

How to understand?

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_command = command.split()
+    if list_command[1] == list_command[2] or list_command[0] != "cp":
```
EdAlekseiev, line 3:
```suggestion
def copy_file(command: str) -> None:
    list_command = command.split()
    cmd, old_file, new_file = list_command
    if old_file == new_file or cmd != "cp":
```

EdAlekseiev (changes requested):

app/main.py
kravetb: Thank you very match!!!!

NikOneZ1 (approved):
Well done

PR #654, opened 2023-10-24
Solution

maksimpro001:


NikOneZ1 (changes requested):
The solution won't work because at the moment it will only copy files with the same name, and should only copy files with different names, as the task description says. Fix that, please

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    text = command.split(" ")
+    if text[0] != "cp":
+        return
+    if text[1] == text[2]:
+        return
```
NikOneZ1, line 6: You able to merge this two conditions
```suggestion
    if text[0] != "cp" or text[1] == text[2]:
        return
```

app/main.py
```diff
@@ -1 +1,8 @@
...
+    with open(text[1], "r") as f1, open(text[2], "w") as f2:
+        f2.write(f1.read())
```
NikOneZ1, line 8: Use more convenient naming 
```suggestion
    with open(text[1], "r") as file1, open(text[2], "w") as file2:
        file2.write(file1.read())
```

NikOneZ1 (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    text = command.split(" ")
```
EdAlekseiev, line 2: `" "` is default value for `split`

```suggestion
    text = command.split()
```

EdAlekseiev (changes requested):

NikOneZ1 (approved):
Well done!

PR #656, opened 2023-10-24
Created file copy program.

AndriiHomeniuk1:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file, new_file = command.split()
```
ArsenPidhoretskyi, line 2: Before you must validate that command.split() have 3 elements
file, new_file bad variable name file != path

app/main.py
```diff
@@ -1 +1,11 @@
...
+
+    if (
+            cp == "cp"
+            and file.endswith(".txt") and len(file.split(".")[0]) > 0
+            and new_file.endswith(".txt") and len(new_file.split(".")[0]) > 0
```
ArsenPidhoretskyi, line 7: Redundant checks

ArsenPidhoretskyi (changes requested):

app/main.py
AndriiHomeniuk1: Yes) i fixed it.

app/main.py
AndriiHomeniuk1: I changed that

app/main.py
```diff
@@ -1 +1,14 @@
...
+        return
+
+    cp, path, new_path = args
+
+    try:
+        with open(path, "r") as file_in, open(new_path, "w") as file_out:
+            file_out.write(file_in.read())
+
+    except FileNotFoundError:
+        print(print(f"File: {path} not found."))
```
vsmutok, line 14: why you need two `print()`?

vsmutok (changes requested):

app/main.py
AndriiHomeniuk1: Oh, I just accidentally added that.

vsmutok (approved):
Great  👍

PR #658, opened 2023-10-25
Solution

DanSheremeta:
Solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    old_file = command.split()[1]
+    new_file = command.split()[2]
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    old_file = command.split()[1]
+    new_file = command.split()[2]
```
EdAlekseiev, line 3: Don't use split twice. You can create a separate variable and store a result of `command.split` in this variable and use it where you need.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #661, opened 2023-10-26
Solution

paashkovaaa:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file, file_copy = command.split()
```
YukihiroSM, line 2: Add check, that command length is 3, cause you might have errors on this step

YukihiroSM (changes requested):
Else looks good!

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if cp != "cp" or file == file_copy or len(command.split()) != 3:
```
YukihiroSM, line 3: It's better to perform command length check before you split the command into variables.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    if len(command.split()) != 3:
+        return
+
+    cp, file, file_copy = command.split()
```
EdAlekseiev, line 6: Don't split string twice. Create a separate variable for storing the result of split and use it where you need.

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #662, opened 2023-10-27
Solution

Artem-user-19:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    # Split the command into parts
```
viktoria-rybenchuk, line 2: remove comment

app/main.py
```diff
@@ -1 +1,10 @@
...
+    command_parts = command.split()
+
+    if command_parts[0] != "cp" or command_parts[1] == command_parts[2]:
+        return
```
viktoria-rybenchuk, line 6: it is better to use unpacking here

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+
+    command1, command2, command3 = command_parts
```
viktoria-rybenchuk, line 4: Check the length of command_parts before unpacking and use descriptive variable names

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if command1 != "cp" or command2 == command3:
+        return
+
+    with (open(command_parts[1], "r") as file_in,
+          open(command_parts[2], "w") as file_out):
```
viktoria-rybenchuk, line 9: You have already variable names for `command_part`. Use them

viktoria-rybenchuk (changes requested):

PR #666, opened 2023-10-31
Solution amen Halilua

hbilous:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[1] == command_list[2]:
```
ArsenPidhoretskyi, line 3: You also need to check than len(command_list) == 3

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
...
+        pass
+    elif command_list[0] == "cp" and len(command_list) == 3:
```
YukihiroSM, line 5: len(command_list) == 3 shoud be done at the very beginning just after split, because if it's less than 3 - you will not get to this line and the script will be crashed ad 3rd line.

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[1] == command_list[2]:
+        pass
```
YukihiroSM, line 4: It should be passes. If files are the same - just stop the function.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[0] == "cp" and len(command_list) == 3:
```
```suggestion
    if len(command_list) == 3 and command_list[0] == "cp":
```
Order make sense because if command_list will be empty it will give an error

ArsenPidhoretskyi (changes requested):

YukihiroSM (approved):
Looks good to me!

PR #667, opened 2023-10-31
Solution for Windows copy

makukh:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    sep_phrases = command.split()
+    old_filename = sep_phrases[1]
+    new_filename = sep_phrases[2]
+    if sep_phrases[0] == "copy" and old_filename != new_filename:
```
EdAlekseiev, line 5: The first element must be `cp`

EdAlekseiev (changes requested):

makukh (commented):
"copy" changed to "cp"

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    sep_phrases = command.split()
```
ArsenPidhoretskyi, line 2:
```suggestion
    parts = command.split()
```
Rename it, because sep_phrases is not clear name

app/main.py
```diff
@@ -1 +1,14 @@
...
+    old_filename = sep_phrases[1]
```
ArsenPidhoretskyi, line 3: Before you need to validate that parts have correct length(3)

ArsenPidhoretskyi (changes requested):

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    old_filename = parts[1]
+    new_filename = parts[2]
```
EdAlekseiev, line 4: Do it after checking length of `parts ` to avoid `IndexError`

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,21 @@
...
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3:
+        print("Invalid command format. Follow this structure:"
+              "cp filename.txt copied_filename.txt")
+        return
+
+    old_filename = parts[1]
+    new_filename = parts[2]
```
EdAlekseiev, line 10: You can use unpacking here

EdAlekseiev, line 10:
```suggestion
    command, old_filename, new_filename = parts
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #668, opened 2023-11-01
Solution

spryima:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file1, file2 = command.split()[1:]
```
vasylhnatiuk, line 2: you must check if command has 3 elements before it

vasylhnatiuk: and also check that command has "cp"

vasylhnatiuk (changes requested):

Grayder0152 (approved):

PR #669, opened 2023-11-01
'Solution'

Illya-Maznitskiy:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3 or parts[0] != "cp":
+        return
+
+    main_file = parts[1]
+    for_copy_file = parts[2]
```
viktoria-rybenchuk, line 8: Consider using unpacking for improved readability and to avoid repeating variable assignments

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3 or parts[0] != "cp":
```
Dimosphen1, line 4: Perform the command check in the second if condition.

app/main.py
```diff
@@ -1 +1,13 @@
...
+        return
+
+    main_file, for_copy_file = parts[1], parts[2]
```
Dimosphen1, line 7: Unpack parts without use of indexes, it will be safe to to so since you checked that there are three elements in the `parts` list
```suggestion
    command_name, main_file, for_copy_file = parts
```

app/main.py
```diff
@@ -1 +1,13 @@
...
+
+    if main_file == for_copy_file:
```
Dimosphen1, line 9: Add check for command here.

Polyakiv-Andrey (approved):
Good job!

PR #671, opened 2023-11-04
Implemented py-copy-file

Aeros72:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_to_copy = command.split(" ")[1]
+    new_file = command.split(" ")[2]
```
vsmutok, line 3: use unpacking  here
also first check command length

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_to_copy = command.split(" ")[1]
```
```suggestion
    file_to_copy = command.split()[1]
```

vsmutok, line 2: Make sure you check that the cp input command is correct.

vsmutok (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #672, opened 2023-11-06
solution

Andrey1104:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def cope_file(command: str) -> None:
+    command_list = command.split(" ")
+    if len(command_list) == 3:
+        with (open(command_list[1], "r") as old,
+              open(command_list[2], "w") as copy):
+            copy.write(old.read())
```
pavlejviki, line 6: you need to check if the command is "cp" and that the destination file is not the same as the original file

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #673, opened 2023-11-06
Solution

AnyoneClown:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_names = command.split()
+    if files_names[1] == files_names[2]:
```
Dimosphen1, line 3: You should additionally check that the `cp` word is present in a command and it is placed in the first place. Additionally, there is a need to check that the input command contains exactly three elements according [to the checklist](https://github.com/AnyoneClown/py-copy-file/blob/develop/checklist.md#code-efficiency). Also, don't use indexes to retrieve elements from the split command, prefer unpacking and writing values to variables after the length check.

app/main.py
```diff
@@ -1 +1,9 @@
...
+        return
+
+    with (open(files_names[1], "r") as main_file,
+          open(files_names[2], "w") as copied_file):
+        for line in main_file:
+            copied_file.write(line)
```
Dimosphen1, line 9: Do not iterate through lines here, you could use `.writelines()` or a combination of `.write()` + `.read()` methods.

Dimosphen1 (changes requested):
Several changes were requested,

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_names = command.split()
+
+    cp_command = files_names[0]
+    first_file = files_names[1]
+    second_file = files_names[2]
+
```
viktoria-rybenchuk, line 7: Extract the command and file names using unpacking. But first check that length command.split is 3

viktoria-rybenchuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #674, opened 2023-11-08
solution

Taran21V:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    sep_command = command.split()
+    if not (len(sep_command) == 3):
+        return
+    cop, file_name, new_file_name = sep_command
+    if file_name == new_file_name:
```
ArsenPidhoretskyi, line 7: also check that command is really cp

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    sep_command = command.split()
+    if not (len(sep_command) == 3):
```
ArsenPidhoretskyi, line 4: not (len(sep_command) == 3) -> len(sep_command) != 3

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Good job

PR #675, opened 2023-11-09
'Solution'

OLENA-KALITSINSKA:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) == 3 and parts[1] != parts[2]:
```
viktoria-rybenchuk, line 4: Make sure you check that the `cp` input command is correct.

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
viktoria-rybenchuk, line 2: You can use tuple unpacking to make your code more readable and to avoid accessing values by index. First, you should check the length of parts to ensure it's 3, and then you can unpack the values into meaningful variable names.

viktoria-rybenchuk: And then you can use variable instead of indexing in all places

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
...
+
+    if parts[0] == "cp" and len(parts) == 3 and parts[1] != parts[2]:
+        source_file, destination_file = parts[1], parts[2]
```
viktoria-rybenchuk, line 5: you still use indexing, first make unpacking like this `cp, source_file, destination_file = parts` and then make another checking with variable names

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    what_command = parts[0]
```
pavlejviki, line 3:
```suggestion
    command = parts[0]
```

app/main.py
```diff
@@ -1 +1,16 @@
...
+    source_file, destination_file = parts[1], parts[2]
```
pavlejviki, line 4: you should do the len check before accessing the parts by index

pavlejviki (changes requested):

PR #676, opened 2023-11-09
Solution

Molodoirockstar:


app/main.py
```diff
@@ -1 +1,17 @@
...
+    if not isinstance(command, str):
+        raise TypeError("Command must be str type")
+    if len(command.split()) != 3 or command.split()[0] != "cp":
+        raise ValueError(
+            "Command should be like this:"
+            " cp file.txt file-copy.txt"
+        )
+
+    first_file_name = command.split()[1]
+    second_file_name = command.split()[2]
```
ArsenPidhoretskyi, line 11:
```suggestion
    first_file_name, second_file_name = command.split()[1:]
```
No need to split so much times
Also better to do split before 
if len(command.split(......

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #677, opened 2023-11-10
'Solution'

Hrechaniuk:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_command = command.split(" ")
+    source_file, destination_file = list_command[1], list_command[2]
+    if source_file != destination_file:
+        with open(source_file, "r") as file_in, \
```
vasylhnatiuk, line 5: try to avoid \

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_command = command.split(" ")
```
```suggestion
    list_command = command.split()
```

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_command = command.split()
+    source_file, destination_file = list_command[1], list_command[2]
```
vasylhnatiuk, line 3: you must check that list_command has 3 arguments before this line. because code can be broken

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #678, opened 2023-11-11
Solution

DenVia:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_names = command.split(" ")
+    if files_names[1] != files_names[2]:
+        with open(files_names[1], "r") as file_to_copy, \
```
LiudmylaKulinchenko, line 4: Use parentheses instead of linebreaker here

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_names = command.split(" ")
```

app/main.py
LiudmylaKulinchenko, line 1: Add a check if command is correct

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None | str:
+    file_names = command.split()
```
LiudmylaKulinchenko, line 2: use unpacking here
here is an example: `command, old_file, new_file = command.split()`

app/main.py
LiudmylaKulinchenko: you should also add check if command is equal to "cp"

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    if len(file_names) != 3 or file_names[1] == file_names[2]:
+        return "invalid format"
```
LiudmylaKulinchenko, line 5: the function should return nothing

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    old_file, new_file = file_names[1], file_names[2]
+
+    # Copy the file content from source to destination
```
LiudmylaKulinchenko, line 9: You can remove this comment, because it is the same as your function name, the logic is clear

app/main.py
```diff
@@ -1 +1,12 @@
...
+    with open(old_file, "r") as file_to_copy, open(new_file, "w") as new_file:
+        file_content = file_to_copy.readlines()
```
LiudmylaKulinchenko, line 11: You can use read() instead of readlines here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, old_file, new_file = command.split()
+    if (
+            old_file == new_file
+            or command != "cp"
+    ):
+        raise Exception
```
LiudmylaKulinchenko, line 7: You can use blank return instead. Raise `Exception` is a bad practice

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, old_file, new_file = command.split()
+    if (
+            old_file == new_file
+            or command != "cp"
+    ):
```
LiudmylaKulinchenko, line 6: no need to split in lines if it fits in one

app/main.py
```diff
@@ -1 +1,11 @@
...
+        raise Exception
+
+    with open(old_file, "r") as file_to_copy, open(new_file, "w") as new_file:
+        file_content = file_to_copy.read()
+        new_file.write("".join(file_content))
```
LiudmylaKulinchenko, line 11: you can remove join now and rewrite 10-11 lines in one line

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
This is, great job 🔥🔥

PR #680, opened 2023-11-11
Solution

CrazyHilly:


app/main.py
```diff
@@ -1 +1,16 @@
...
+    try:
+        command_type, input_file_name, output_file_name = command.split(" ")
+    except ValueError:
+        print("Correct input format: <command> <input file name> <output file name>")
+        return
+    except FileNotFoundError:
+        print("Input file not found")
+        return
+
+    if input_file_name == output_file_name:
```
ArsenPidhoretskyi, line 11: Also check that command cp

ArsenPidhoretskyi (changes requested):

app/main.py
CrazyHilly: I got an error while uploading, so fixed that along with other changes, and force-pushed the commit.
I forgot that it's being sent for review despite the errors and only saw your comment after everything was done.
Could you please review it from the beginning? Sorry about that.

ArsenPidhoretskyi (approved):
Well done!

PR #681, opened 2023-11-12
Solution

yul1valko:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    all_data = command.split(" ")
```
Dimosphen1, line 2: There is no need to specify that the delimiter is space in `split()` it is a default value.

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    all_data = command.split()
+
+    if all_data[0] != "cp" or all_data[1] == all_data[2] or len(all_data) != 3:
+        return
```
```suggestion
    if len(all_data) != 3 or all_data[0] != "cp" or all_data[1] == all_data[2]:
        raise Exception("some informative  message ")
```

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    all_data = command.split()
```
vsmutok, line 2: I think `all_data`  bad  variable name in this case

app/main.py
```diff
@@ -1 +1,11 @@
...
+
+    if all_data[0] != "cp" or all_data[1] == all_data[2] or len(all_data) != 3:
+        return
+
+    file_to_copy = all_data[1]
+    new_file = all_data[2]
```
```suggestion
    _, file_to_copy, new_file = all_data
```

use unpacking

vsmutok (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #683, opened 2023-11-15
Solution

OleksandrMV:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split()
```
NikOneZ1, line 2: It is not a good naming, because you will have also a command name in the "files", that's not obvious

NikOneZ1 (changes requested):
Add a check that file names are not equal, and that command is "cp"

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command.split()
+    if file_names[1] != file_names[2]:
```
YukihiroSM, line 3: What if length of file_names is less than 3?

YukihiroSM: Also, check that command is cp.

YukihiroSM (changes requested):
+ Fix flake

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command.split()
+    if (file_names[1] != file_names[2]
+            and len(file_names) == 3
```
EdAlekseiev, line 4: You must check length of `file_names` firstly.

EdAlekseiev (changes requested):

NikOneZ1 (approved):
Well done!

PR #686, opened 2023-11-17
Solution

mrve4nick:


app/test.txt
```diff
@@ -0,0 +1,12 @@
...
+asjdflasjdv
+jkavadfvjkdfvdlkf;jv
+dv
+adfv
+fdv
+
+dfv
+fdv
+d
+f
```
pavlejviki, line 12: remove this file

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split(" ")
```
pavlejviki, line 2: you need to make sure that the command is "cp" and that the length of command_split is 3

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if command_split[1] != command_split[2]:
```
pavlejviki, line 3: use unpacking and save the command parts in variables

pavlejviki (changes requested):

pavlejviki (approved):
Good job!

PR #687, opened 2023-11-19
Solution

lyaskov:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, old_file, new_file = command.split(" ")
```
EdAlekseiev, line 2: `" "` is the default value for `split`
```suggestion
    command, old_file, new_file = command.split()
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #689, opened 2023-11-21
Solution

Evheniia96:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    part = command.split()
+    if not part[0].startswith("cp"):
```
EdAlekseiev, line 3: Also check that `part` contains 3 elements
```suggestion
    if part[0] != "cp":
```

EdAlekseiev (approved):
Good job

PR #690, opened 2023-11-22
Solution

Marinel444:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file, new_file = command.split(" ", maxsplit=3)
```
pavlejviki, line 2: Setting maxsplit=2 means the string will be split into at most three parts, which aligns with your expectation.

pavlejviki: but he maxsplit parameter specifies the maximum number of splits, not the exact number. If the actual number of possible splits in the string is less than maxsplit, the method will perform as many splits as possible.

pavlejviki: so its better to perform an additional check using len()  to make sure that the command has exactly three parts

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split(" ")) == 3:
```
Dimosphen1, line 2: There is no need to specify that `split` should be performed over space `" "`, it is a default argument for `split`. Remove `" "`.

app/main.py
```diff
@@ -1 +1,8 @@
...
+        command, file, new_file = command.split(" ")
```
Dimosphen1, line 3: There is no need to perform a `split` operation twice, make it once and write an obtained list to a variable that you will use for length checking and unpacking.

app/main.py
```diff
@@ -1 +1,8 @@
...
+        if command == "cp":
+            if file != new_file:
```
Dimosphen1, line 5: There is no need to make nested `if` conditions here, write the overall condition in one `if` statement.

app/main.py
```diff
@@ -1 +1,8 @@
...
+                with open(file, "r") as file_in, \
+                        open(new_file, "w") as file_out:
```
Dimosphen1, line 7: Avoid using line breaking operator `\` it is a better practice to place conditions inside of `()`, e.g.:
```suggestion
                with (open(file, "r") as file_in,
                      open(new_file, "w") as file_out):
```
Apply this change to your code.

app/main.py
Dimosphen1, line 3: It would be better to not override the `command` variable that comes to the function. You could allocate another variable for the command keyword, e.g. `command_keyword`.

Dimosphen1 (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #691, opened 2023-11-22
'Solution'

SerhiiDomik:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split()
+    source_file = commands[1]
```
vasylhnatiuk, line 3: What if command is []? Check if command has 3 arguments

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    commands = command.split()
+    if len(commands) == 3 and commands[0] == "cp":
+        source_file = commands[1]
+        destination_file = commands[2]
```
vsmutok, line 6: better to use unpacking 
```suggestion
        _, source_file, destination_file = commands
```

vsmutok (changes requested):

vsmutok (approved):
Good Job 👍

PR #692, opened 2023-11-23
Copy-File-Solution

MykolaMazh:


app/main.py
```diff
@@ -1 +1,10 @@
...
+def copy_file(command: str) -> None:
+    arguments = command.split()
+    if (command.startswith("cp ")
+            and len(arguments) == 3
+            and arguments[1] != arguments[2]):
+        file1 = open(arguments[1], "r")
+        file2 = open(arguments[2], "w")
+        file2.write(file1.read())
+        file1.close()
+        file2.close()
```
viktoria-rybenchuk, line 10: Using the `with` statement is recommended when working with file operations in Python because it ensures that the file is properly closed, even if an exception occurs during the execution of the block

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    arguments = command.split()
+    if (command.startswith("cp ")
```
viktoria-rybenchuk, line 3: you should check that command exactly is `cp` not just start with this letters

viktoria-rybenchuk: Unpack the source and destination file paths for clarity

viktoria-rybenchuk (changes requested):

app/main.py
MykolaMazh: Does unpack mean to create the variables?

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_elements = command.split()
+    _command = command_elements[0]
+    file_to_copy = command_elements[1]
+    file_to_write = command_elements[2]
```
viktoria-rybenchuk, line 5:
```suggestion
    _command, file_to_copy, file_to_write = command_elements
```

viktoria-rybenchuk: Extract elements from the list returned by command.split() and assign them to variables.

viktoria-rybenchuk: You should check the length of elements before unpacking to avoid unexpected errors

viktoria-rybenchuk (changes requested):

viktoria-rybenchuk (changes requested):
Fix according to comments

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_elements = command.split()
+    _command = command_elements[0]
+    if len(command_elements) != 3 or _command != "cp":
```
```suggestion
    if len(command_elements) != 3 or command_elements[0] != "cp":
```

vsmutok (changes requested):

vsmutok (approved):
Great  👍

PR #693, opened 2023-11-24
Solution

Liliyakhu:
variant with os using google

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,25 @@
-# write your code here
+import os
+
+
+# variant 1
+def copy_file_1(command: str) -> None:
+    splited_command = command.split(" ")
+    if splited_command[1] == splited_command[2]:
+        return
+    # src = "file.txt"
+    # dest = "new_file.txt"
+    os.system(command)
+
+
+# variant 2
+def copy_file_2(command: str) -> None:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,25 @@
...
+    splited = command.split(" ")
+    source = splited[1]
+    destination = splited[2]
+    if source == destination:
+        return
+    with open(source, "r") as file_in, open(destination, "w") as file_out:
+        text_to_copy = file_in.read()
+        file_out.write(text_to_copy)
+
+# there can be a third variant that uses shutil module
```
viktoria-rybenchuk, line 25: remove all redundant comments

viktoria-rybenchuk (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, source, destination = command.split(" ")
```
viktoria-rybenchuk, line 2: Recommendation: Use `split` with whitespace as the default separator. First, check the length of the command list before unpacking to avoid unexpected errors.

viktoria-rybenchuk (approved):
Good job!

PR #695, opened 2023-11-27
Solution

dkibalenko:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split(" ")
+    if split_command[1] != split_command[2]:
```
pavlejviki, line 3: Make sure you check that the `cp ` input command is correct.

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split(" ")
```
Dimosphen1, line 2: There is no need to explicitly indicate that the split symbol should be space, it is a default option.

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if len(split_command) == 3:
+        if (split_command[0] == "cp"
+                and split_command[1] != split_command[2]):
+            with (open(split_command[1], "r") as file_in,
+                  open(split_command[2], "w") as file_out):
```
Dimosphen1, line 7: Instead of retrieving different elements after split by indexing, you could unpack `split_command` and write arguments to variables thus making your code more descriptive. Change your approach to use unpacking.

app/main.py
```diff
@@ -1 +1,9 @@
...
+                content = file_in.read()
+                file_out.write(content)
```
Dimosphen1, line 9: There is no need for the `content` variable, pass `file_in.read()` in the `file_out.write()` directly.

Dimosphen1 (changes requested):

viktoria-rybenchuk (approved):
Well done!

PR #697, opened 2023-11-28
Solution

Vlad1slavo:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        print("Invalid command")
+        return
+    source, destination = parts[1], parts[2]
```
```suggestion
    source, destination = parts[1:]
```

ArsenPidhoretskyi (approved):
Small fix) Better to use unpacking

ArsenPidhoretskyi (changes requested):
.

PR #700, opened 2023-11-29
Solution_29112023

Anton-Konyk:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
```
ArsenPidhoretskyi, line 2: Rename command_list -> parts
_list suffix is redundant

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if command_list[0] == "cp" and command_list[1] != command_list[2]:
```
ArsenPidhoretskyi, line 3: Also check that length of parts is equal to 3

ArsenPidhoretskyi (changes requested):

NikOneZ1 (approved):
Well done!

PR #703, opened 2023-12-01
Solution

Pavlo-here:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_of_tokens = command.split()
+    if list_of_tokens[1] == list_of_tokens[2]:
```
ArsenPidhoretskyi, line 3: Also check that tokens length is equal to 3 and command is really cp

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    tokens = command.split()
+    if len(tokens) != 3 and tokens[0] != "cp" and tokens[1] == tokens[2]:
```
vsmutok, line 3:
```suggestion
    if len(tokens) == 3 and tokens[0] != "cp" and tokens[1] == tokens[2]:
```

vsmutok (changes requested):

vsmutok (approved):
Well Done!

PR #706, opened 2023-12-06
Solution

Ded31:


EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(copy_command: str) -> None:
+    copy_commands = copy_command.split()
+    old_name = copy_commands[1]
+    new_name = copy_commands[2]
```
NikOneZ1, line 4: You can simplify this
```suggestion
    command, old_name, new_name = copy_command.split()
```

NikOneZ1 (changes requested):
Check if the command is cp and that command contains three parameters

EdAlekseiev (approved):
Good job

PR #708, opened 2023-12-08
'Solution'

zakotii:


EdAlekseiev (changes requested):

YukihiroSM (approved):
Looks good! 
Also, you can check that parts[0] equals "cp"

PR #710, opened 2023-12-09
Solution

Serhii-Hrynyshyn:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command[3:].split(" ")
+    f_in = str(command[0])
+    f_out = str(command[1])
```
YukihiroSM, line 4: Why are you ignoring first 3 characters of command? It's better to first split the command -> check that first element of command is command name (cp), and check that command consists of 3 parts only. Please, add these checks

YukihiroSM (changes requested):

NikOneZ1 (approved):
Well done!

PR #713, opened 2023-12-15
added copy_file func

BGBroth:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, fname_to_copy, new_fname = command.split(" ")
```
pavlejviki, line 2: check that the splitted command consists of three parts before unpacking

app/main.py
```diff
@@ -1 +1,11 @@
...
+
+    if not command == "cp":
+        pass
+
+    if fname_to_copy == new_fname:
+        pass
```
pavlejviki, line 8: you can join this two if conditions with `and` and better raise an informative warning instead of pass

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_arguments = command.split(" ")
+    if len(splitted_arguments) != 3:
+        raise ValueError("Incorrect comand provided")
+    command, fname_to_copy, new_fname = command.split(" ")
```
Dimosphen1, line 5: There is no need to perform an additional split, you could unpack the `splitted_arguments` variable.

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_arguments = command.split(" ")
```
Dimosphen1, line 2: There is no need to indicate that you should split on space, it is a default option. Leave just `command.split()`.

app/main.py
```diff
@@ -1 +1,14 @@
...
+    if len(splitted_arguments) != 3:
+        raise ValueError("Incorrect comand provided")
+    command, fname_to_copy, new_fname = command.split(" ")
+
+    if not command == "cp":
+        raise ValueError("Incorrect comand provided")
+
+    if fname_to_copy == new_fname:
+        raise ValueError("Incorrect comand provided")
```
Dimosphen1, line 11: Combine these two checks into one to reduce the number of if conditions.

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_arguments = command.split(" ")
+    if len(splitted_arguments) != 3:
+        raise ValueError("Incorrect comand provided")
+    command, fname_to_copy, new_fname = command.split(" ")
+
+    if not command == "cp":
```
Dimosphen1, line 7: It would be preferable to change this line to
```suggestion
    if command != "cp":
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+        raise ValueError("Incorrect comand provided")
```
Dimosphen1, line 8: Change the typo in the `comand` word to `command`.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #714, opened 2023-12-16
ready!

job-artem:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_1, file_2 = command.split(" ")
```
pavlejviki, line 2: check that the splitted command consists of three parts before unpacking and that the command is "cp"

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp_command, file_1, file_2 = command.split(" ")
```
```suggestion
    cp_command, file_1, file_2 = command.split()
```

vsmutok: check that the splitted command consists of three parts before unpacking

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if file_1 == file_2 or cp_command != "cp":
+        pass
+    else:
+        with open(file_1, "r") as source, open(file_2, "a") as target:
+            for line in source:
+                target.write(line)
```
```suggestion
        target.write(source.read())
```

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp_command, file_1, file_2 = command.split(" ")
+    if file_1 == file_2 or cp_command != "cp":
+        pass
```
vsmutok, line 4: better to raise some Exeptions

vsmutok: so that it is clear to the user if something goes wrong

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+class WrongArgumentsNumberError(Exception):
+    """Raise when user command consists not from 3 elements"""
+
+
+class WrongCommandLengthError(Exception):
+    """Raise when user command have less than 6 symbols"""
+
+
+def copy_file(command: str) -> None:
+    if len(command) < 6:
```
pavlejviki, line 18: Since you are validating that the command contains exactly three parts by splitting the string and checking its length, additional validation of the overall command length is unnecessary. This three-part check is more relevant for your purpose, as it directly ensures that the command follows the expected format, regardless of the individual lengths of the command and file names.

app/main.py
```diff
@@ -1 +1,34 @@
...
+                                      "to make copying possible, example:"
+                                      "cp 1 2")
+    cp_command, file_1, file_2 = command.split(" ")
+    if file_1 == file_2:
+        raise FileCopyError("Source and target file are the same!")
+    if cp_command != "cp":
+        raise WrongCommandError("You written wrong command, "
+                                "if you wanna copy file, use cp command")
+    if len(command.split(" ")) != 3:
+        raise WrongArgumentsNumberError("Your copy command "
```
pavlejviki, line 30: this check should be done prior to unpacking

pavlejviki (changes requested):

PR #715, opened 2023-12-16
feat: Solution

sashokpilipok7:


app/main.py
```diff
@@ -1 +1,16 @@
...
+
+
+def get_full_path(file_name: str) -> str:
+    dir_path = os.path.dirname(os.path.abspath(__file__))
+    return os.path.join(dir_path, file_name)
+
+
+def copy_file(command: str) -> None:
+    arguments = command.split(" ")
+    if (arguments[1] != arguments[2]):
```
vsmutok, line 11: Make sure you check that the cp input command is correct.

vsmutok: also check if arguments length is 3

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #717, opened 2023-12-17
Solution

Vityk-4W:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    res = command.split()
+    if res[1] != res[2]:
```
EdAlekseiev, line 3: Also check, that the first element is `cp`

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #718, opened 2023-12-19
Copy file

Ananiev-Vitalii:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_r = command.split(" ")[1]
+    file_w = command.split(" ")[2]
```
EdAlekseiev, line 3: Don't call split twice. You can create separate variable for storing result of `split ` and use if where you need it. Also, `" "` is the default value for split
```suggestion
    file_r = command.split()[1]
    file_w = command.split()[2]
```

EdAlekseiev (changes requested):

PR #719, opened 2023-12-19
solve py-copy-file

zdmytryszyn:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    file_to_copy_name = command[1]
+    new_file_name = command[2]
```
pavlejviki, line 4: Make sure you check that the cp input command is correct.

Check the length of the splitted command to make sure it consists of three parts. Better use unpacking instead of indexing

pavlejviki (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
```
pavlejviki, line 2: Check the length of the splitted command to make sure it consists of three parts

pavlejviki (changes requested):

PR #725, opened 2023-12-26
Solution

KwintDimon:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cd, source_name, copy_name = command.split()
```
Dimosphen1, line 2: To perform a safe unpacking independently from the number of elements after split, you can extract exactly the first three elements and neglect others using the following syntaxis:
```suggestion
    cd, source_name, copy_name, *_ = command.split()
```

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if source_name == copy_name:
```
Dimosphen1, line 3: Perform an additional check that `cd` is `"cp"` and if not, return.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cd, source_name, copy_name, *_ = command.split()
+    if cd != "cp":
+        return
+    if source_name == copy_name:
```
```suggestion
    if source_name == copy_name or cd != "cp":
```

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cd, source_name, copy_name, *_ = command.split()
+    if cd != "cp":
```
vasylhnatiuk, line 3: and delete this

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #727, opened 2024-01-02
Implemented "copy_file" function. However, I haven't found any tests in the forked directory for some reason...

raychw:
Implemented "copy_file" function. However, I haven't found any tests in the forked directory for some reason...

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    key_elements = command.split()
+
+    if key_elements[0] != "cp":
+        raise ValueError("Invalid command. Maybe you meant 'cp...'?")
+    if key_elements[1] == key_elements[2]:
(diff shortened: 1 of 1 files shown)


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    key_elements = command.split()
```
```suggestion
   command_arguments= command.split()
```

app/main.py
```diff
@@ -1 +1,11 @@
...
+
+    if key_elements[0] != "cp":
+        raise ValueError("Invalid command. Maybe you meant 'cp...'?")
+    if key_elements[1] == key_elements[2]:
```
vasylhnatiuk, line 6: you must check if command has 3 argumenta before it.

vasylhnatiuk (changes requested):

PR #734, opened 2024-01-08
Solution

Zonda001:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split()
+    file_name_1 = command_split[1]
+    file_name_2 = command_split[2]
```
vsmutok, line 4: You must check if command has 3 argumenta before it.
Also check, that command start with cp

vsmutok (changes requested):

vsmutok (approved):
Well Done!

PR #740, opened 2024-01-10
Solution

dariiashostak:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    ls_command = command.split(" ")
```
```suggestion
    ls_command = command.split()
```

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if ls_command[1] != ls_command[2] and ls_command[0] == "cp":
+        with (open(ls_command[1], "r") as source,
+              open(ls_command[2], "w") as new_file):
```
vsmutok, line 5: it's better to unpack values into the variables, like: `var, var, var = parts`. Then use these variables instead of indexes, it will increase readability

vsmutok (changes requested):

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #742, opened 2024-01-11
Solution

kate-chern:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file_from, file_to = command.split(" ")
```

vasylhnatiuk (changes requested):

PR #743, opened 2024-01-11
Solution

maxbidun:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(text: str) -> None:
+    if len(text.split()) == 3:
+        text = text.split()
+        if text[0] == "cp" and text[1] != text[2]:
+            with open(text[1], "r") as file_in, open(text[2], "w") as file_out:
```
vsmutok, line 5: it's better to unpack values into the variables, like: var, var, var = parts. Then use these variables instead of indexes, it will increase readability

vsmutok (changes requested):

PR #751, opened 2024-01-15
Solution

Acik33:


app/file.txt
```diff
@@ -0,0 +1,2 @@
+Hello Arsen
+Do your code work ?
```
vsmutok, line 2: remove app/file.txtle from PR

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split(" ")
```
```suggestion
    files = command.split()
```

app/main.py
```diff
@@ -1 +1,8 @@
...
+    old_file = files[1]
+    new_file = files[2]
+
+    if old_file != new_file:
```
vsmutok, line 6: you should check if command is exactly cp. Read checklist

vsmutok (changes requested):

PR #754, opened 2024-01-17
Solution

MaxymChyncha:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_str_into_list = command.split()
+    old_file, new_file = command_str_into_list[1], command_str_into_list[2]
```
aLEKS-e3, line 3: you can write this in 1 line -> `command, old_file, new_file= command.split()` because there are only 3 elements in your `command_str_into_list`.

aLEKS-e3 (changes requested):

PR #755, opened 2024-01-17
Solution

angelzagniy:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command[3:].split()
```
vasylhnatiuk, line 2: what if command has less then 3 arguments? add check for this case

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,13 +1,11 @@
 def copy_file(command: str) -> None:
-    if len(command) > 3:
-        _, *file_names = command.split()
-        if (
-                command[0:2] == "cp"
-                and command.count(" ") == 2
-                and file_names[0] != file_names[1]
+    cp, source, destination = command.split()
```
vasylhnatiuk, line 2: add if len(command) >= 3: before it

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #756, opened 2024-01-18
Develop

PyCore23:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+from os import path, listdir, getcwd, chdir
+from sys import argv
+
+
+def copy_file(command: str) -> None:
+    chdir(path.dirname(argv[0]))
+    ref = command.split(" ")[1]
```
vasylhnatiuk, line 7: what is ref? what if command does not have 3 arguments?

app/main.py
```diff
@@ -1 +1,11 @@
...
+    new = command.split(" ")[2]
```
vasylhnatiuk, line 8: use only one time split(), don't repeat yourself

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+from os import path, chdir
+from sys import argv
+
+
+def copy_file(command: str) -> None:
+    parts = command.split(" ")
```
EdAlekseiev, line 6: `" "` is default value for `split`.
```suggestion
    parts = command.split()
```

app/main.py
```diff
@@ -1 +1,17 @@
...
+    if len(parts) < 3:
```
EdAlekseiev, line 7: Also, check that command start with `cp`

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split(" ")
```
```suggestion
    parts = command.split()
```

vsmutok: `" "` is default value for `split()`

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if len(parts) < 3:
```
vsmutok, line 3: Also, check that command start with `cp`

app/main.py
```diff
@@ -1 +1,13 @@
...
+        print("invalid command format")
+    else:
+        source = parts[1]
+        new = parts[2]
```
```suggestion
        _, source, new = parts
```

vsmutok (changes requested):

app/main.py
vsmutok: `" "` is default value for `split()`

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split(" ")
+    if len(parts) < 3 or parts[1] == parts[2]:
+        raise ValueError("Invalid command")
+
+    try:
+        with open(parts[1], "r") as source_file:
+            with open(parts[2], "w") as destination_file:
```
vsmutok, line 8: use one `with`

```python
with open(.., "r") as f, open(..., "w") as f2:
```

vsmutok (changes requested):

PR #758, opened 2024-01-18
Implement copy-file task

Dmytry-S:


NazarSheremet (changes requested):
At the bottom of the task there is a Note block, which contains additional points specifying that it is also necessary to check if the first `str=="cp"`

Dmytry-S:
Updated

YuriiSmolii (approved):
Nice

PR #759, opened 2024-01-18
Solution

u123dev:


app/example.tst
```diff
@@ -0,0 +1,11 @@
+def copy_file(cp_cmd: str) -> None:
```
vasylhnatiuk, line 1: why do you nedd this file?

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(cp_cmd: str) -> None:
+    arg = cp_cmd.split(" ")
+    if arg[0] != "cp" or arg[1] == arg[2]:
+        return
+    with open(arg[1], "r") as fin, open(arg[2], "w") as fout:
```
vasylhnatiuk, line 5: fin, fout are bad names

app/main.py
```diff
@@ -1 +1,11 @@
...
+        content = fin.read()
+        fout.write(content)
+
+
+# =====
+copy_file("cp main.py example.tst")
```
vasylhnatiuk, line 11: you don't need to call function. delete this

vasylhnatiuk (changes requested):

PR #761, opened 2024-01-19
Solution

IvanGuculyak:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_vile(command: str) -> None:
+    text = command.split()
+    if text[0] == "cp" and text[1] != text[2]:
```
vsmutok, line 3: First of all, you should check that command.split() have exactly 3 parameters

vsmutok: it's better to unpack values into the variables, like: var, var, var = parts. Then use these variables instead of indexes, it will increase readability

vsmutok (changes requested):

PR #762, opened 2024-01-20
Solution

Meksim-fesh:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    instruction, f_in_name, f_out_name = command.split()
+    if instruction != "cp" or f_in_name == f_out_name:
```
vsmutok, line 3: it is better to raise informative error so that it is clear to the user if something goes wrong

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    instruction, f_in_name, f_out_name = command.split()
```
vsmutok, line 2: before unpacking you should check that command.split() has 3 arguments

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #764, opened 2024-01-20
Solution

barabarinov:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
MaxymChyncha, line 2: You can create variables directly on this line for later use. Like this `command, source_file, copied_file = command.split()`

MaxymChyncha (approved):
Good job! 
Check my comment.

app/main.py
```diff
@@ -1,15 +1,12 @@
...
-    if len(parts) == 3 and parts[0] == "cp":
-        source_file, copied_file = parts[1], parts[2]
-
-        if source_file != copied_file:
-            with (
-                open(source_file, "r") as read_file,
-                open(copied_file, "w") as write_file,
-            ):
-                write_file.write(read_file.read())
+    _, source_file, copied_file = command.split()
```
NEkropo1, line 2: what if `_ != "cp"`?)

NEkropo1 (changes requested):

app/main.py
barabarinov: > what if `_ != "cp"`?)

But we don`t use "cp" here. Why we need to save it in variable?

NEkropo1: exactly what I said. Why we don't use it?)

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if source_file != copied_file:
+        with (
+            open(source_file) as read_file,
+            open(copied_file, "w") as write_file,
+        ):
+            write_file.write(read_file.read())
+
+
+if __name__ == "__main__":
+    copy_file("cp test.txt new_test.txt")
```
NEkropo1, line 13:
```suggestion
    copy_file("hehehaha test.txt new_test.txt")
```
I mean this command should not work )

NEkropo1 (changes requested):

PR #771, opened 2024-01-23
Solution

oksankakot:


yulias97 (approved):
I like your code! Good job!

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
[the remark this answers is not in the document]
oksankakot: Fixed code according to your remarks

app/main.py
```diff
@@ -1,10 +1,9 @@
 def copy_file(command: str) -> None:
     try:
-        cp, main_file, for_copy_file = command.split()
+        command, main_file, for_copy_file = command.split()
     except ValueError:
         return
-
-    if main_file == for_copy_file:
+    if command != "cp" or main_file == for_copy_file:
         return
```
NEkropo1, line 7:
```suggestion
if command == "cp" and main_file == for_copy_file:
    <your other code>
```

Try to not finish code like this, I mean raw returns are not a good sign

NEkropo1 (changes requested):

app/main.py
oksankakot: Fixed that

PR #772, opened 2024-01-23
Solution

yulias97:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    filenames = command.split()
+    source_filename = filenames[1]
+    destination_filename = filenames[2]
```
barabarinov, line 7: You can do this in one line! ;)

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import shutil
```
barabarinov, line 1: I think you don`t need that! Just use .write() and .read()

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+
+def copy_file(command: str) -> None:
+    filenames = command.split()
+    source_filename = filenames[1]
+    destination_filename = filenames[2]
+    if source_filename == destination_filename:
+        return
+    with open(source_filename, "r") as file_in, \
```
barabarinov, line 10: You can delete "r" mode, because it goes by default.

barabarinov (approved):

app/main.py
yulias97: I think write() and .read() is not as effective as using a ready-made method

app/main.py
```diff
@@ -1,12 +1,6 @@
-import shutil
-
-
 def copy_file(command: str) -> None:
-    filenames = command.split()
-    source_filename = filenames[1]
-    destination_filename = filenames[2]
-    if source_filename == destination_filename:
+    origin_name, copy_name = command.split()[1], command.split()[2]
+    if origin_name == copy_name:
```
NEkropo1, line 3: what if command is wrong?

NEkropo1 (changes requested):

app/main.py
```diff
@@ -1,6 +1,6 @@
 def copy_file(command: str) -> None:
-    origin_name, copy_name = command.split()[1], command.split()[2]
-    if origin_name == copy_name:
+    cp, origin_name, copy_name = command.split()
+    if cp != "cp" or origin_name == copy_name:
         return
```
NEkropo1, line 4: one more task, try to change your code so you won't get raw return, because it's a bad sign

NEkropo1 (changes requested):

app/main.py
yulias97: ok
I removed `return`

NEkropo1 (approved):
Great work!

PR #773, opened 2024-01-23
Solution

Anna728560:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_in, file_out = command.split()
```
Nick-r2c, line 2: Make sure you check that the `cp` input command is correct.  It might be 'int' or start with, let's say, 'foo' instead of 'cp')

Nick-r2c (approved):

Serhii-Leonenko (changes requested):

Y-Havryliv (approved):
gj!

PR #779, opened 2024-01-24
'Solution'

NataliiaKorzh:


app/main.py
```diff
@@ -1 +1,26 @@
...
+
+        while True:
+
+            if parts[1] == parts[2]:
+                print("Copy cannot have the same name!")
+                break
+
+            if len(parts) != 3 or parts[0] != "cp":
+                print("Error: Invalid command format. "
+                      "Please use 'cp file copy_file'")
```
isikach, line 19: As for me it looks better if you check it before you open file

app/main.py
```diff
@@ -1 +1,26 @@
...
+                return
```
isikach, line 20: You can use raise here as well

isikach (approved):

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+    with (open(parts[1], "r") as file_origin,
+          open(parts[2], "w") as file_copy):
+
+        text = file_origin.read()
+
+        while True:
+
+            if parts[1] == parts[2]:
+                raise NameError("Copy cannot have the same name!")
```
NEkropo1, line 17: I know it's intuitive to use NameError here, but it's wrong, you should create custom error, something like SameFileNamesError, since default name error in python used in different scenarios.

NEkropo1 (changes requested):

PR #781, opened 2024-01-24
Solution

isikach:


artemgrishko (approved):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> None:
+    linux_command, source_name, new_file = command.split(" ")
+    if (
+            linux_command != "cp" or not (
```
NEkropo1, line 4: better just command. We're supposed to use it cross-platform, not only on linux, so not much sense in adding linux_

app/main.py
```diff
@@ -1 +1,19 @@
...
+            source_name.endswith(".txt")
+            or new_file.endswith(".txt")
```
NEkropo1, line 6: what if we want to copy content not from txt file?

app/main.py
```diff
@@ -1 +1,19 @@
...
+            )):
+
+        raise ValueError("Check your command and try again")
+    if source_name == new_file:
+        raise ValueError("Names should not have the same name")
```
NEkropo1, line 11: actually in task was written: does nothing. So, probably, it should do nothing, not Raise an error, even though it's intuitive.

app/main.py
```diff
@@ -1 +1,19 @@
...
+    try:
+        source = open(source_name, "r")
+    except (FileNotFoundError, IOError):
+        print("File not found")
```
NEkropo1, line 15: why do we need this if we anyway will throw that error?

app/main.py
```diff
@@ -1 +1,19 @@
...
+    else:
+        with source as file_in, open(new_file, "w") as file_out:
+            file_out.write(file_in.read())
+        print(f"File {source_name} copied to {new_file} successfully")
```
NEkropo1, line 19: linux cp doesn't provide any feedback, so we don't need this print either

NEkropo1, line 19: linux cp command doesn't provide any feedback, so we don't need it either

NEkropo1 (changes requested):

app/main.py
isikach: My version is more userfriendly)

PR #782, opened 2024-01-24
Solution

kstorozhenko:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    if command_parts[0] != "cp" or command_parts[1] == command_parts[2]:
+        print("Command is not 'cp' or source "
```
shaiduchyk, line 4: imho u can use variables, like source/destination_file or smth, not with index. I think it will be look better.

shaiduchyk (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp_command, source, destination_file = command.split()
+    if cp_command != "cp" or source == destination_file:
+        print("Command is not 'cp' or source "
+              "and destination files are the same.")
+        return
```
NEkropo1, line 6: raise an error here, or refactor logic, way we don't have raw return.

NEkropo1 (changes requested):
https://github.com/mate-academy/py-copy-file/pull/782#discussion_r1465979318
still not fixed.

kstorozhenko:
> [#782 (comment)](https://github.com/mate-academy/py-copy-file/pull/782#discussion_r1465290989) [#782 (comment)](https://github.com/mate-academy/py-copy-file/pull/782#discussion_r1465979318) still not fixed.

sry, apparently when I was testing the code's functionality I added this again and forgot to remove it
now if fix it)

app/main.py
```diff
@@ -7,6 +7,3 @@ def copy_file(command: str) -> None:
     with open(source, "r") as file_in, open(destination_file, "w") as file_out:
         file_out.write(file_in.read())
     print(f"Copied contents from {source} to {destination_file}")
```
NEkropo1, line 9: linux cp command also is not giving any feedback, so this print is also excessive)

NEkropo1 (changes requested):

app/main.py
kstorozhenko: done)

NEkropo1 (approved):
Good job!

PR #787, opened 2024-01-24
'Solution'

artemlitvenko:


app/main.py
```diff
@@ -1 +1,13 @@
...
+    parts = command.split()
+
+    if len(parts) == 3 and parts[0] == "cp":
+        source_file, destination_file = parts[1], parts[2]
+
+        if source_file == destination_file:
+            return
+
+        with open(source_file, "r") as file_in:
+            with open(destination_file, "w") as file_out:
```
motivated2, line 12: You can use two contexts managers simultaneously by separating them by a comma, there is an example in README

motivated2 (changes requested):

PR #790, opened 2024-01-24
Solution

AntonSkyrda:


Dobryk12 (approved):

app/main.py
```diff
@@ -1 +1,28 @@
...
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3:
+        print("Invalid command."
+              "Please provide a command in the format: "
+              "cp source_file destination_file")
+        return
+
+    _, source_file, destination_file = parts
```
NEkropo1, line 10: what if I use it like "delete_both_files file1 file2"? always read checklist

app/main.py
```diff
@@ -1 +1,28 @@
...
+
+    if source_file == destination_file:
+        print("Source and destination file names are the same."
+              "Nothing will be done.")
+        return
```
NEkropo1, line 15: raise an error or refactor logic, don't use raw returns

app/main.py
```diff
@@ -1 +1,28 @@
...
+              open(destination_file, "w") as file_out):
+            file_content = file_in.read()
+            file_out.write(file_content)
+
+        print(f"File '{source_file}' copied to "
+              f"'{destination_file}' successfully.")
+    except FileNotFoundError:
+        print(f"Error: File '{source_file}' not found.")
+    except Exception as error:
+        print(f"Error: {error}")
```
NEkropo1, line 28: better simply check if file exists, and leave error raising to native python, without try except for no reason

NEkropo1 (changes requested):

PR #792, opened 2024-01-28
Solution

DenZatolokin:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[1] == command_list[2]:
```
vasylhnatiuk, line 3: you must check if command list has 3 arguments before this line

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
...
+        return
+    if command_list == "cp" and len(command_list) == 3:
```
YukihiroSM, line 5: command_list is a list and can't be compared to string in this case. Please, fix it. You should check the zero element of list. 
Also, the check should be performed before checking files identity as @vasylhnatiuk previously mentioned

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, file_copy, file_new = command.split()
```
EdAlekseiev, line 2: you must check if command list has 3 arguments before this line

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #793, opened 2024-01-29
'Solution'

sofiazholud:


app/main.py
```diff
@@ -1 +1,24 @@
...
+        return
+
+    # Copy the file content
+    with open(source_file, "r") as file_in:
+        content = file_in.read()
+    with open(destination_file, "w") as file_out:
+        file_out.write(content)
+
+
+copy_file("cp source_file.txt destination_file.txt")
```
YukihiroSM, line 24: This should be removed, or transformed into: 
```suggestion
if __name__ == "__main__":
    copy_file("cp source_file.txt destination_file.txt")
```

That's because if you will import this module somewhere it will start working immideately now, but you don't need it.

app/main.py
```diff
@@ -1 +1,24 @@
...
+
+    # If the file names are the same, do nothing
+    if source_file == destination_file:
+        return
+
+    # Copy the file content
+    with open(source_file, "r") as file_in:
+        content = file_in.read()
+    with open(destination_file, "w") as file_out:
+        file_out.write(content)
```
YukihiroSM, line 21: What about using single `with`? 

`with(open(source_file, "r") as file_in, open(destination_file, "w") as file_out):
...
`

YukihiroSM (changes requested):
Also, source and destination files shouldn't be in the repo :)

Else looks good. Fix comments, please.

app/main.py
sofiazholud: changed it.

app/main.py
sofiazholud: changed it.

EdAlekseiev (approved):
Good job

PR #795, opened 2024-01-29
solution

OlgaRozd:


app/main.py
```diff
@@ -1 +1,13 @@
...
+
+    if source_file == destination_file:
+        print("Source and destination files are the same.")
+    try:
+        with (open(source_file, "r") as file_in,
+              open(destination_file, "w") as file_out):
+            file_out.write(file_in.read())
+    finally:
+        file_in.close()
+        file_out.close()
```
Dimosphen1, line 13: - You should check that the command contains the `cp` part;
- Do not explicitly close files, it would be done automatically while exiting the context manager;
- Do not unpack args using indexes, you could use the following approach:
` command_name, source_file, destination_file, *_ = args`

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    args = command.split()
+    if args[0] != "cp":
+        raise ValueError("Command must start with 'cp'.")
+    command_name, source_file, destination_file, *_ = args
+    if source_file == destination_file:
+        raise ValueError("Source and destination files are the same.")
```
Dimosphen1, line 7: The better ordering for operations would be the following
```suggestion
    args = command.split()
    command_name, source_file, destination_file, *_ = args
    
    if command_name != "cp":
        raise ValueError("Command must start with 'cp'.")
    if source_file == destination_file:
        raise ValueError("Source and destination files are the same.")
```
Use an approach that allows you to avoid using indexing and replace it with descriptive variable names where possible.

Dimosphen1 (approved):
LGTM, one comment is left for consideration.

PR #797, opened 2024-01-30
Solution

ViaMal:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[1] != command_list[2]:
```
vasylhnatiuk, line 3: you must check if command_list has 3 arguments

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if command_list[1] != command_list[2] and len(command_list) == 3:
```
```suggestion
    if len(command_list) == 3 and command_list[1] != command_list[2] :
```

vasylhnatiuk: change order

vasylhnatiuk (changes requested):

PR #798, opened 2024-01-31
Solution

mykytaso:


vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp_command, source_name, destination_name = command.split()
```
vsmutok, line 2: before unpacking, you should check that the length of `command.split()` is three

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if source_name != destination_name and cp_command == "cp":
+        print("1")
```
vsmutok, line 4: why???

vsmutok (changes requested):

app/main.py
mykyta-so: sorry, I forgot to delete it

ArsenPidhoretskyi (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #800, opened 2024-02-01
Solution

cth-usq:


app/main.py
```diff
@@ -1 +1,10 @@
...
+def copy_file(command: str) -> None:
+    name = command.split(" ")
+
+    if len(name) == 3 and name[0] == "cp" and not name[1] == name[2]:
+
+        with open(name[1], "r") as file_in:
+            text = file_in.read()
+
+        with open(name[2], "w") as file_out:
+            file_out.write(text)
```
NazarSheremet, line 10: This part of code can be written better
```suggestion
         with (open(copies_file, "r") as file_in,
                  open(result_file, "w") as file_out):
                file_out.write(file_in.read())
```

NazarSheremet (changes requested):
Minor chanches

app/main.py
cth-usq: Fixed

YuriiSmolii (approved):
LGTM

PR #809, opened 2024-02-04
Solution

AlexGrytsai:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands_list = command.split()
+    copy_command = commands_list[0]
+    old_file = commands_list[1]
+    new_file = commands_list[2]
```
EdAlekseiev, line 5: You should check that `commands_list` contains 3 elements before unpacking values

EdAlekseiev, line 5: Use unpacking here
```suggestion
    copy_command, old_file, new_file = commands_list
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if copy_command != "cp":
+        raise NameError(f"{copy_command} is not a valid command.")
+    if old_file == new_file:
+        raise ValueError(f"File {old_file} and {new_file} have same name")
+    with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+        for line in file_in:
+            file_out.write(line)
```
EdAlekseiev, line 12: You don't need file here
```suggestion
        file_out.write(file_in.read())
```

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #810, opened 2024-02-04
task completed

voronvasia:


app/main.py
```diff
@@ -1 +1,10 @@
...
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if (len(parts) != 3
+            or parts[0] != "cp"
+            or parts[1] == parts[2]):
+        return
+    with (open(parts[1], "r") as file_in,
+          open(parts[2], "w") as file_out):
+        file_in_read = "".join(file_in.read())
+        file_out.write(str(file_in_read))
```
Dimosphen1, line 10: There is no need to create a `file_in_read` variable and cast string to string several times here. You could pass read directly to write, e.g.:
```suggestion
        file_out.write(file_in.read())
```
Rewrite the code to follow this approach.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #812, opened 2024-02-05
Solution

voidmain69:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    components = command.split()
+    command_, source_file, destination_file = command.split()
```
NikOneZ1, line 3: You can use components instead of command.split() here
```suggestion
    command_, source_file, destination_file = components
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+    if len(components) == 3 and command_ == "cp":
+        if source_file != destination_file:
```
NikOneZ1, line 6: You able to merge this if statements
```suggestion
    if len(components) == 3 and command_ == "cp" and source_file != destination_file:
```

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #814, opened 2024-02-06
Solution

halyna-baklanova:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> bool:
+    parts = command.split()
+
+    source_file = parts[1]
+    destination_file = parts[2]
+
+    if source_file == destination_file:
```
Dimosphen1, line 7: You should also check that the command is actually `"cp"` here, in case not, you could return.

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> bool:
+    parts = command.split()
+
+    source_file = parts[1]
+    destination_file = parts[2]
```
Dimosphen1, line 5: Tru to avoid retrieving elements by index and use unpacking instead
```suggestion
    command_name, source_file, destination_file, *_ = command.split()
```

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> bool:
```
Dimosphen1, line 1: Change your return type annotation, there is no `bool` value returned.

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
...
+    command_name, source_file, destination_file, *_ = command.split()
+
+    if command_name != "cp":
+        return False
+
+    if source_file == destination_file:
```
pavlejviki, line 7: use the `or` operator to combine these conditions

pavlejviki (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #816, opened 2024-02-08
Solution

MykhailoIvchenko:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split(" ")
```
```suggestion
    command_parts = command.split()
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    if not command_parts[0] or command_parts[0] != "cp":
```
EdAlekseiev, line 4: You should check that `command_parts` contains 3 elements

EdAlekseiev (changes requested):

app/main.py
MykhailoIvchenko: Fixed

app/main.py
MykhailoIvchenko: Done

EdAlekseiev (approved):
Good job

PR #819, opened 2024-02-11
function to copy file using linux command

RomanHlodann:


app/main.py
```diff
@@ -1 +1,11 @@
...
+def copy_file(command: str) -> None:
+    arguments = command.split()
+    if len(arguments) != 3:
+        raise ValueError("Wrong count of arguments")
+    com, original_filename, copy_filename = arguments
+    if com != "cp":
+        raise ValueError("Wrong command")
+    with (open(arguments[1], "r") as file_in,
+          open(arguments[2], "w") as file_out):
+        if original_filename != copy_filename:
```
Dimosphen1, line 10: It would be better to place this check in the if condition on line 6 to not open files without necessity. Place this check earlier.

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    arguments = command.split()
+    if len(arguments) != 3:
+        raise ValueError("Wrong count of arguments")
+    com, original_filename, copy_filename = arguments
+    if com != "cp":
+        raise ValueError("Wrong command")
+    with (open(arguments[1], "r") as file_in,
+          open(arguments[2], "w") as file_out):
```
Dimosphen1, line 9: Do not use indexing, you have these values in the variables.

Dimosphen1 (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #820, opened 2024-02-11
solved task

Shantagust:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.startswith("cp") and len(command.split()) == 3:
```
Dimosphen1, line 2: This check is not legit since `cp` is not the whole command name, it could be `cpath` or another command name that should not be considered valid. Change the code to check that the command is exactly `cp`.

Dimosphen1 (changes requested):

PR #821, opened 2024-02-11
added solution

MAFIOZNIK1999:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[0] == "cp" and command[1] != command[2]:
```
LiudmylaKulinchenko, line 3: you can unpack these variables in the easiest way, also use the more descriptive names

app/main.py
```diff
@@ -1 +1,11 @@
...
+        with open(command[1], "r") as file_input:
+            with open(command[2], "w") as file_output:
```
LiudmylaKulinchenko, line 5: You can nest the context managers using a single with statement and separating them with commas to achieve a one-liner for opening both files

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good!

PR #822, opened 2024-02-11
Solution

vsupruniuk:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, input_name, output_name = command.split(" ")
```
LiudmylaKulinchenko, line 2: no need to use space inside split method, it uses space by default

LiudmylaKulinchenko (approved):
Well done 🔥

PR #828, opened 2024-02-18
Solution

LitvinOleksandr:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(files_name: str) -> None:
+    files_name = files_name.split(" ")
+    if len(files_name) == 3 \
+            and files_name[1] != files_name[2] \
+            and files_name[0] == "cp":
```
EdAlekseiev, line 5: Use `()` instead of `\`

app/main.py
```diff
@@ -1 +1,8 @@
...
+        with open(str(files_name[1]), "r") as file_read, \
+                open(str(files_name[2]), "w") as file_write:
```
EdAlekseiev, line 7: The same comment as above

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #831, opened 2024-02-19
Solution

kolya2731:


app/main.py
```diff
@@ -1 +1,10 @@
...
+def copy_file(command: str) -> None:
+    if (
+            len(command.split()) == 3
+            and command.split()[0] == "cp"
+            and command.split()[1] != command.split()[2]
+    ):
+
+        with (open(command.split()[1], "r") as file,
+              open(command.split()[2], "w") as new_file):
+            new_file.write(file.read())
```
Dimosphen1, line 10: Make the `split` operation once and assign output to a variable afterward. Check the length of the obtained variable and in case it equals 3 perform unpacking and use newly created variables during two checks that are left. Avoid using indexing

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    com, source_file, target_file = (command_parts[0],
```
```suggestion
    command, source_file, target_file = command.split()
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+                                     command_parts[1],
+                                     command_parts[2])
+    if (
+            com == "cp"
+            and source_file != target_file
+            and len(command_parts) == 3
```
vasylhnatiuk, line 9: you should do this check before 3 line

vasylhnatiuk (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #832, opened 2024-02-20
'Solution'

Yevhen23tk:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    split_com = command.split()
+
+    cp = split_com[0]
+    source_file = split_com[1]
+    destination_file = split_com[2]
```
NikOneZ1, line 7: You can use unpacking here
```suggestion
    cp, source_file, destination_file = split_com
```

NikOneZ1 (changes requested):

app/main.py
Yevhen23tk: ok, accepted

NikOneZ1 (approved):

PR #833, opened 2024-02-21
SOL

neagarr:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.split()[0] == "cp":
```
vasylhnatiuk, line 2: use command.split() only one time

app/main.py
```diff
@@ -1 +1,8 @@
...
+
+        with open(f"{command.split()[1]}", "r") as file,\
```
vasylhnatiuk, line 4: use () instead of \

vasylhnatiuk (changes requested):

neagarr:
can't use () instead of \

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    if element[0] == "cp":
+        with open(
+                f"{element[1]}", "r"
+        ) as file, open(
+            f"{element[2]}", "a"
+        ) as file_copy:
+
+            if file.name != file_copy.name:
+                file_copy.write(f"{file.read()}")
```
vsmutok, line 12: no need to use `f-string` here

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    element = command.split()
```
vsmutok, line 2: element is not clear name

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_part = command.split()
```
```suggestion
    command_parts = command.split()
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    if command_part[0] == "cp":
+        with open(
+                command_part[1], "r"
+        ) as file, open(
+            command_part[2], "a"
+        ) as file_copy:
```
vsmutok, line 9:
```suggestion
        with (
            open(command_parts[1], "r") as file, 
            open( command_parts[2], "a") as file_copy
        ):
```

vsmutok (changes requested):

app/main.py
neagarr: when I try to use parentheses I get a syntax error

neagarr: изображение

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+
+    if command_parts[0] == "cp" and len(command_parts) == 3:
```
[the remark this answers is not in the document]
neagarr: Thanks

app/main.py
vsmutok: @neagarr 
image

vsmutok: @neagarr maybe you get syntax error because you use old python
I approved this task, but for the future try to use newer versions of python
image

vsmutok (approved):

PR #836, opened 2024-02-22
'Solution'

donbas777:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    sort_command = command.split()
```
Dimosphen1, line 2: Use unpacking after the length check here to avoid using indexes and be more descriptive

app/main.py
```diff
@@ -1 +1,5 @@
...
+    with (open(sort_command[1], "r") as file_in,
+          open(sort_command[2], "w") as file_out):
+        file_out.write(f"{file_in}")
```
Dimosphen1, line 5: Instead of f-string use .`read()` method
```suggestion
        file_out.write(file_in.read())
```

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    sort_command = command.split()
+    with (open(sort_command[1], "r") as file_in,
```
Dimosphen1, line 3: Before opening files make sure that the incoming and outcoming paths are not the same and the command name is actually `cp`

Dimosphen1 (changes requested):

PR #838, opened 2024-02-23
Solution

IhorDmytriv:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    params = command.split(" ")
+    command_name = params[0]
+    file_name = params[1]
+    file_name_copy = params[2]
```
NikOneZ1, line 5: You can use unpacking here, but first of all check if there is 3 params in command

NikOneZ1 (changes requested):

vasylhnatiuk (approved):

PR #839, opened 2024-02-24
Solution

MaksymNikulyak:


app/main.py
```diff
@@ -1 +1,19 @@
...
+            command_tokens.remove("")
+
+    if len(command_tokens) != 3:
+        raise SyntaxError("File-copy command must consist of three tokens")
+    if command_tokens[0] != "cp":
+        raise SyntaxError("File-copy command is not recognized")
+    if command_tokens[1] == command_tokens[2]:
+        return
+
+    with open(command_tokens[1], "r") as file_in, \
```
vasylhnatiuk, line 16: use () instead of \

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command_name: str) -> None:
+    command_name = command_name.lower()
+    command_tokens = command_name.split(" ")
```
```suggestion
    command_tokens = command_name.split()
```

vasylhnatiuk: you can avoid it

vasylhnatiuk (changes requested):

app/main.py
MaksymNikulyak: Thanks!

app/main.py
MaksymNikulyak: Sorry, but the parentheses give me an error -- please, see the attached picture
скрін_6

MaksymNikulyak: -- PyCharm says: "Unexpected expression syntax".

Could it be because of my obsolete Python version? -- Have 3.8

EdAlekseiev: Problem in your version of Python. You can use parentheses only from python 3.10

EdAlekseiev (changes requested):

app/main.py
MaksymNikulyak: I would have gladly upgraded but my PyCharm's version is 2020.1.5 and it only supports the Python versions up to, and including, 3.9.

So, kindly, am unable to avoid the awkward syntax in this case

EdAlekseiev (approved):
Good job

PR #845, opened 2024-02-26
Solution

RomanNest:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> object:
+    new_str = command.split(" ")
```
```suggestion
    new_str = command.split()
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+    file_current = str(new_str[1])
+    file_future = str(new_str[2])
```
EdAlekseiev, line 4: You should check, that `new_str` contains 3 elements

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #846, opened 2024-02-27
solution

rtsadm:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_in = command.split()[1]
```
vasylhnatiuk, line 2: use command.split() only one time, and you must check that command is `cp` and there are exactly 3 arguments

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split = command.split()
+    if split[1] != split[2] and split[0] == "cp" and len(split) == 3:
```
```suggestion
    if len(split) == 3 and split[1] != split[2] and split[0] == "cp" :
```

vasylhnatiuk: order is important in this case

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #848, opened 2024-02-28
solution

vanko6226:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    c_arr = command.rsplit()
```
EdAlekseiev, line 2: You can use `split`
```suggestion
    c_arr = command.split()
```

app/main.py
```diff
@@ -1 +1,5 @@
...
+    if c_arr[0] == "cp" and (c_arr[1] != c_arr[2]):
```
EdAlekseiev, line 3: Firstly check that `c_arr` contains 3 elements

EdAlekseiev (changes requested):

PR #853, opened 2024-03-06
'Solution'

sdannyk:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    old_file = parts[1]
+    new_file = parts[2]
```
NikOneZ1, line 4: Try to use unpacking here, but before check if the command has 3 parameters and if the command if equal to cp

NikOneZ1 (changes requested):

EdAlekseiev (approved):
Good job

PR #859, opened 2024-03-10
'Solution'

olessh:


app/main.py
```diff
@@ -1 +1,13 @@
...
+        raise ValueError
+
+    if command_ls[1] != command_ls[2]:
+        with (open(command_ls[1], "r") as file_in,
+              open(command_ls[2], "w") as file_out):
+            data = file_in.read()
+            file_out.write(data)
+
+    else:
+        pass
```
vsmutok, line 13: why you need it???

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_ls = command.split()
+    if command_ls[0] != "cp":
+        raise ValueError
+
+    if command_ls[1] != command_ls[2]:
```
vsmutok, line 6: You should check, that command_ls contains 3 elements

vsmutok (changes requested):

PR #866, opened 2024-03-13
solution

kostiukmkalne:
I`m not sure about my tests
I saw I had some trouble

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file = command.split()
+
+    if file == 3:
+        if file[1] != file[2] and file[0] == "cp":
+            with (
+                open(command[1], "r") as file_in,
+                open(command[2], "w") as file_out
+            ):
+                file_out.write(file_in.read())


kostiukmkalne:
I do not understand what`s wrong with my tests?!

kostiukmkalne:
I changed name
This is so bad practice call file like <file>🤪

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    new_file = command.split()
+
+    if new_file == 3:
+        if new_file[1] != new_file[2] and new_file[0] == "cp":
```
NazarSheremet, line 5: Why need to do 2 conditions? Merge them

NazarSheremet (changes requested):
Need changes

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    new_file = command.split()
+
+    if new_file == 3:
```
NazarSheremet, line 4: what does it mean `new_file == 3`? may it be?
```suggestion
    if len(new_file) == 3:
```

NazarSheremet (approved):
Good job!

PR #867, opened 2024-03-13
Solution

comnotaDidi:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def cope_file(command: str) -> any:
+    new_command = command.split()
+    if new_command[1] == new_command[2]:
```
pavlopro, line 3: Please add an additional check to ensure that the length of new_command is 3 and that the zeroth element is 'cp'.

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def cope_file(command: str) -> any:
```
pavlopro, line 1: Please change the return annotation to `None`

pavlopro (changes requested):

pavlopro (approved):
Good job!

PR #868, opened 2024-03-13
Done

antonver:


app/main.py
```diff
@@ -1 +1,7 @@
 # write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
```
Dimosphen1, line 3: Perform length check and use unpacking to avoid using indexing in your code

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if command[1] == command[2]:
```
Dimosphen1, line 4: Make sure that the command is actually `cp` before performing further actions

Dimosphen1 (changes requested):

PR #869, opened 2024-03-13
Solution

SHCHERBANV:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split(" ")
```
NazarSheremet, line 2: by default `split()` function split by space, so no reason to write it 
```suggestion
    commands = command.split()
```

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if commands[0] == "cp":
+        if commands[1] != commands[2]:
```
NazarSheremet, line 4: Why need to do 2 conditions? Merge them

NazarSheremet (changes requested):
Need small changes

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split(" ")
```
YuriiSmolii, line 2: Take a look on previous Nazar's comment and refactor your split command.

YuriiSmolii (changes requested):
Minor change requested.

PR #870, opened 2024-03-13
Solution

Viesich:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_origin = command.split()[1]
+    cp_file = command.split()[2]
```
Dimosphen1, line 3: Use unpacking and do not perform split operation twice. Before unpacking check that the number of elements after split equals to three

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if file_origin != cp_file:
```
Dimosphen1, line 4: Make sure that the command is actually `cp` before performing file operation

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    com, file_origin, cp_file = parts
+    if len(parts) == 3 and com == "cp" and file_origin != cp_file:
```
vasylhnatiuk, line 4: you should check `if len(parts) == 3` before     `com, file_origin, cp_file = parts`

vasylhnatiuk (changes requested):

PR #876, opened 2024-03-18
Solution

rosttv:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    read_file = command.split()[1]
+    write_file = command.split()[2]
+    if read_file != write_file and command.split()[0] == "cp":
```
Dimosphen1, line 4: Perform a `split` operation once and write it to a variable, after that check the length of the variable, and in case its length is 3 use unpacking and later use obtained variables in the logic. Do not retrieve values index here

app/main.py
```diff
@@ -1 +1,10 @@
...
+        with (
+            open(read_file, "r") as file_in,
+            open(write_file, "w") as file_out
+        ):
+            content = file_in.read()
+            file_out.write(content)
```
Dimosphen1, line 10: There is no need to create a `content` variable, pass `file_in.read()` directly to the `file_out.write(` instead

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+
+    if len(command) == 3:
```
pavlopro, line 4: Good. Now you can unpack into variables and make the check, as you already did in the previous commit.

pavlopro (changes requested):

app/main.py
rosttv: Ok, but I don't understand why you need to check for length == 3? everything else was done in the last commit

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    key_word, read_file, write_file, *_ = command.split()
+    """# After use "*_" instruction no need "len == 3" verif,
```
pavlopro, line 3: Imagine that the `command `is "cp test". You will encounter an error in this case.

pavlopro (changes requested):

app/main.py
rosttv: I agree, but in this case it is more appropriate to catch ValueError. I don't like the check for len == 3 precisely from the point of view of style, because this way you have to duplicate the operation "command.split()"

pavlopro: You don't need to duplicate the command.split() operation. You can do it like this:
....
parts = command.split()
if len(parts) == 3:
    command, read_file, write_file = parts
....

pavlopro (changes requested):
Please check my last comment

Polyakiv-Andrey (approved):
Good job!

PR #878, opened 2024-03-19
Solution

Bujhm99:


Boyok1337 (approved):
LGTM

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    name_massive = command.split(" ")
```
YuriiKindrat, line 2: no need touse indexes here, u could unpack ur vars like this:
`command, file_in, file_out = command.split()`
also there is no need to pass space as arg in split method, it use space by default

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if name_massive[1] != name_massive[2]:
```
YuriiKindrat, line 3: also add a check for command, what if command will be not 'cp' ?

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ!

PR #879, opened 2024-03-19
Solution

Boyok1337:


Bujhm99 (approved):
LGTM!

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file, copied_file = command.split(" ")
```
YuriiKindrat, line 2: there is no need to pass space as arg in split method, it use space by default

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if file != copied_file:
```
YuriiKindrat, line 3: make a check for `cp` command, what if command will be not `cp`?

app/main.py
```diff
@@ -1 +1,7 @@
...
+        with (open(file, "r") as reading_file,
+              open(copied_file, "w") as writing_file):
+            reading_file_data = reading_file.read()
+            writing_file.write(reading_file_data)
```
YuriiKindrat, line 7: could u do this in one line ?

YuriiKindrat (changes requested):

Sofbog (approved):
LGTM

PR #880, opened 2024-03-19
'Solution'

Vitalii-Haponiuk:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    old_file, new_file = (command[2:].split())
```
pavlopro, line 2: There is no guarantee that the first two characters are 'cp'. So, please simply split the command and also check that the length of the split result equals 3. 
Additionally check that fist element is 'cp'

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if old_file != new_file:
+        with open(old_file, "r") as file:
```
pavlopro, line 4: Please try to use context managers on a single line

pavlopro (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command.split()
```
pavlopro, line 2: Check if length of `file_names` is 3

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if file_names[1] != file_names[2] and file_names[0] == "cp":
```
pavlopro, line 3: Unpack file_names into a variables for better readability.

pavlopro (changes requested):

Dimosphen1 (approved):
LGTM, well done!

PR #883, opened 2024-03-20
Solution?

OKochubeii:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split(" ")
```
NazarSheremet, line 2: By default split() function split by space, so no reason to write it
```suggestion
    command_split = command.split()
```

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if command_split[0] != "cp":
+        return
+    file_name = command_split[1]
+    file_copy = command_split[2]
```
NazarSheremet, line 6: It can be done simpler, but also need to add a check if the command contains 3 parts
`command_name, file_name, file_copy= command.split(" ")`

NazarSheremet (changes requested):

app/main.py
OKochubeii: Дякую за відгук, вже виправив🫡

NazarSheremet (approved):
Good job!

PR #886, opened 2024-03-21
Solution

Vova-Stoliar:


app/main.py
```diff
@@ -1 +1,21 @@
...
+    if len(normalized_command) != copy_command_words_total:
+        raise Exception(copy_command_error_message)
+
+    copy_command, file_to_copy_path, destination_path = normalized_command
+
+    if copy_command != standard_copy_command:
+        raise Exception(copy_command_error_message)
+
+    if file_to_copy_path == destination_path:
+        return
```
LiudmylaKulinchenko, line 17: Don't use a lot of `if` statements.

LiudmylaKulinchenko, line 17: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+def copy_file(command: str) -> None:
+    standard_copy_command = "cp"
+    copy_command_words_total = 3
+    copy_command_error_message = "Bad Command"
+
+    normalized_command = command.split(" ")
```
LiudmylaKulinchenko, line 6: no need to use space inside split method, it uses space by default

LiudmylaKulinchenko (changes requested):

app/main.py
Vova-Stoliar: I do not think there is a need to do this as it increases nesting

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+def copy_file(command: str) -> None:
+    standard_copy_command = "cp"
+    copy_command_words_total = 3
+    copy_command_error_message = "Bad Command"
+
+    normalized_command = command.split()
```
LiudmylaKulinchenko, line 6: you can unpack these variables in the easiest way, also use the more descriptive names

app/main.py
```diff
@@ -1 +1,21 @@
...
+    is_copy_command_valid = normalized_command[0] != standard_copy_command
+    is_command_valid = len(normalized_command) != copy_command_words_total
+
+    if is_command_valid or is_copy_command_valid:
+        raise Exception(copy_command_error_message)
+
+    file_to_copy_path, destination_path = normalized_command[1:]
+
+    if file_to_copy_path == destination_path:
+        return
```
LiudmylaKulinchenko, line 17: Why don't just make it one condition? You can also then use one backward condition to avoid blank return
I'd also suggest not to change error message, because it won't be descriptive

LiudmylaKulinchenko (changes requested):

PR #887, opened 2024-03-21
'Solution'

o-zinevych:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command.split(" ")
```
```suggestion
    file_names = command.split()
```

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if file_names[1] != file_names[2]:
```
EdAlekseiev, line 3: Also, check that `file_names ` contains 3 elements and first element is "cp"

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #889, opened 2024-03-24
solution

TymurSh:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    ls = command.split(" ")
```
vasylhnatiuk, line 2: ls is a bad name

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if ls[1] == ls[2]:
```

vasylhnatiuk (changes requested):

app/main.py
```suggestion
    ls = command.split()
```

vasylhnatiuk (changes requested):

app/main.py
TymurSh: Why?

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split()
+    if len(files) == 3:
```
```suggestion
    if len(files) != 3 or files[1] == files[2]:
        return 
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split()
+    if len(files) != 3 or files[1] == files[2]:
```
EdAlekseiev, line 3: You should check that command start with "cp"

EdAlekseiev (changes requested):

app/file.txt
vsmutok, line 1: remove this file from PR

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split()
+    if len(files) != 3 or files[1] == files[2] or files[0] != "cp":
+        return
```
vsmutok, line 4: better print some message or raise Exception  so that if something goes wrong, the user can understand it

vsmutok (changes requested):

vasylhnatiuk (approved):

vasylhnatiuk (approved):

PR #890, opened 2024-03-24
Solution

TetyanaPavlyuk:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(input_str: str) -> None:
+    command, file_name, new_file_name = input_str.split()
```

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #897, opened 2024-03-25
Solution

Line181:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(text: str) -> None:
+    if text.split() == 3:
+        command, file_name_1, file_name_2 = text.split()
```
EdAlekseiev, line 3: Don't use `split` twice. You should create separate variable for storing result of `spilt` and use it where you need

app/main.py
```diff
@@ -1 +1,7 @@
...
+        if command == "cd" and file_name_1 != file_name_2:
+            with open(file_name_1, "r") as source, \
+                    open(file_name_2, "w") as copy:
```
EdAlekseiev, line 6: Use `()` for separating lines instead of `\`

EdAlekseiev (changes requested):

EdAlekseiev (approved):
Good job

PR #898, opened 2024-03-27
Solution

Sergunshot:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def coopy_file(command: str) -> None:
+    elements = command.split(" ")
```
EdAlekseiev, line 2: `" "` is default value
```suggestion
    elements = command.split()
```

app/main.py
```diff
@@ -1 +1,16 @@
...
+    source_file, destination_file = elements[1], elements[2]
+
+    if source_file != destination_file:
+        with open(source_file, "r") as file_in, open(destination_file,
+                                                     "w") as file_out:
+            file_out.write(file_in.read())
+            print(f"File {file_in} copied to {file_out}."
+                  f" Successfully")
+    else:
+        pass
```
EdAlekseiev, line 16:
```suggestion
```

EdAlekseiev (changes requested):

app/main.py
Sergunshot: Done

EdAlekseiev (approved):
Good job

PR #901, opened 2024-03-28
Solution

vitalinamal:


Bujhm99 (approved):
LGTM, but no need to use indexes here, u could unpack ur vars like this:
command, file_in, file_out = command.split()

also add a check for command, what if command will be not 'cp' ?

vitalinamal:
> LGTM, but no need to use indexes here, u could unpack ur vars like this: command, file_in, file_out = command.split()
> 
> also add a check for command, what if command will be not 'cp' ?

I did an additional check
When I was doing the task, I thought about writing this check, but I remembered when the mentors corrected me because I did additional checks for tasks that did not require it (exactly the same as this task)

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split()
+    if len(commands) != 3 or commands[0] != "cp" or commands[1] == commands[2]:
+        return
+
+    _, copied_file, new_file = commands
+    with open(copied_file, "r") as c_file, open(new_file, "w") as n_file:
```
Oleksl888, line 7: You can add error handling here in case file path does not exist

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split()
+    if len(commands) == 3:
+        cp, copied_file, new_file = commands
+        if cp == "cp" and copied_file != new_file:
+            with (open(copied_file, "r") as c_file,
```
Oleksl888, line 6: you can use built-in functions to simplify this

app/main.py
```diff
@@ -1 +1,8 @@
...
+                  open(new_file, "w") as n_file):
```
Oleksl888, line 7: So how does your code handle file names that do not exist or will create an exception?

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Good job 🌼

PR #902, opened 2024-03-29
created function which copy data from file to another file

include031:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = [file for file in command.split(" ")]
```
Polyakiv-Andrey, line 2: Here you don't need to use a list generator to get a list of commands, just write Command.split().

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if command_list[1] != command_list[2]:
```
Polyakiv-Andrey, line 3: add a check that the first command is “cp”

Polyakiv-Andrey (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #904, opened 2024-03-29
Solution

panicua:


oleksandra-shershen (approved):
GJ!

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
```
Oleksl888, line 2: you can use split() without specifying a delimiter - it will automatically split by whitespace

app/main.py
```diff
@@ -1 +1,8 @@
...
+    first_file, second_file = command_list[1], command_list[2]
```
Oleksl888, line 3: what about first element of the command? what if it doesn't equal 'cp'? What if command contains more then 3 words?

app/main.py
```diff
@@ -1 +1,8 @@
...
+
+    if first_file != second_file:
+        with (open(first_file, "r") as file_from,
```
Oleksl888, line 6: You can simplify this by using commands from built in `shutil` library.
Also how does your code handle invalid input (non-existing files, invalid file names)?

Oleksl888 (changes requested):

app/main.py
panicua: > what about first element of the command? what if it doesn't equal 'cp'? What if command contains more then 3 words?

That's the question for the people who made testes for the task, but I'll fix it.

app/main.py
panicua: > You can simplify this by using commands from built in `shutil` library. Also how does your code handle invalid input (non-existing files, invalid file names)?

I'll checkout shutil lib after the morning lesson, and I add try/except to handle invalid inputs if shutil doesn't do it also.

app/main.py
Oleksl888: In real life when you expect user input be ready for anything :)

LiudmylaKulinchenko: @panicua You have this as first point in your [checklist](https://github.com/mate-academy/py-copy-file/blob/master/checklist.md) 👀`Code Efficiency Make sure you check that the cp input command is correct.`

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None | str:
```
LiudmylaKulinchenko, line 4: There is no option that your function will return None. Fix annotations, plese

app/main.py
```diff
@@ -1 +1,20 @@
...
+    try:
+        shutil.copy(first_file, second_file)
+        return f"File '{second_file}' was created"
+    except FileNotFoundError:
+        return f"File '{first_file}' doesn't exist"
+    except OSError:
+        return f"Invalid file name '{second_file}'"
+    except Exception:
+        """Incase some unknown error still exist that I don't know of"""
+        raise Exception
```
LiudmylaKulinchenko, line 20: You removed default error message here. You shouldn't do that

app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None | str:
+    command_list = command.split()
+    if command_list[0] == "cp" and len(command_list) == 3:
+        _, first_file, second_file = command_list
+    else:
+        return f"Wrong command line please double check it: {command}"
```
LiudmylaKulinchenko, line 9: You can raise exception here with custom message
Same for other return statements

LiudmylaKulinchenko (changes requested):

oleksandra-shershen (approved):
LGTM!

LiudmylaKulinchenko (approved):
Well done 😌

PR #905, opened 2024-03-30
Solution for task copy file

oleksandra-shershen:


mikhailyemets (approved):
KRASOTA!

app/main.py
```diff
@@ -1 +1,45 @@
...
+    command_name, source_file, destination_file = commands
+
+    if command_name != "cp":
+        raise InvalidCommand("Invalid command. Use 'cp' to copy files.")
+
+    if source_file == destination_file:
+        raise SameFileError("Source file and destination file are the same.")
+
+    try:
+        with (open(source_file, "r") as file_in,
```
Oleksl888, line 37: This is the only piece of code where you really need exceptions. OSError should cover all cases if you try to copy an invalid / non-existing file.
Also you can simplify this by using `shutil` library

app/main.py
```diff
@@ -1 +1,45 @@
-# write your code here
+class InvalidCommandFormat(Exception):
```
Oleksl888, line 1: Appreciate the effort but you don't need most of these. 
Also  it is a good practice to keep exceptions in separate module and import from there

app/main.py
```diff
@@ -1 +1,45 @@
...
+    if len(commands) != 3:
+        raise InvalidCommandFormat("Invalid command format. "
+                                   "Usage: cp source_file destination_file")
+
+    command_name, source_file, destination_file = commands
+
+    if command_name != "cp":
+        raise InvalidCommand("Invalid command. Use 'cp' to copy files.")
+
+    if source_file == destination_file:
```
Oleksl888, line 33: `It must do nothing in case the user is trying to copy file to file with the same name.`
this contradicts the task requirement

Oleksl888 (changes requested):

app/main.py
YuriiKindrat: it is better to use backward condition here, instead of blank return

YuriiKindrat (changes requested):

PR #906, opened 2024-03-30
copy file

mikhailyemets:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split(" ")
+    if files[1] != files[2]:
```
y-kondrashova, line 3: How does your code handle a case if there's wrong command, not cp

y-kondrashova (changes requested):

app/main.py
mikhailyemets: Didn't admit that it was necessary. But there's np to implement stuff like this.

mikhailyemets: "Make sure you check that the `cp` input command is correct."
You're right, btw.)

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None | str:
+    if len(command.split(" ")) < 3:
```
YuriiKindrat, line 2: no need to pass space inside split, it is using it by default

app/main.py
```diff
@@ -1 +1,13 @@
...
+        return "Invalid command format"
+
+    cp_command, input_file_name, output_file_name = command.split(" ")
```
YuriiKindrat, line 5: same

YuriiKindrat (changes requested):

fsocie7y (approved):
well done ^_^

PR #909, opened 2024-03-31
Develop

DmytroHlazyrin:


panicua (approved):
try/except for this case probably is overintended cause in our tests we don't use "copy" / cp.
Plus you have to write "Exception" that is a bad practice. But overall everything is good.

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        copy, file_in, file_out = command.split(" ")
```
YuriiKindrat, line 3: no need to pass space as arg here, split uses it by default

YuriiKindrat: also it is better to name command as 'command' not copy )

YuriiKindrat (changes requested):

fsocie7y (approved):
GJ 💯

PR #910, opened 2024-03-31
Solution

D-u-x:


app/main.py
```diff
@@ -1 +1,15 @@
...
+    try:
+        copy, old_file, new_file = command.split()
+
+        if old_file != new_file or copy == "copy":
+            with (
+                open(old_file, "r") as file_int,
+                open(new_file, "w") as file_out
+            ):
+                file_out.write(file_int.read())
+    except Exception:
```
vasylhnatiuk, line 11: you should write more clear exception, what if Exception connected with file , not with command?(for example - file doesn't exist). you should write another message in this case

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        copy, old_file, new_file = command.split()
+
+        if old_file != new_file or copy == "copy":
```
EdAlekseiev, line 5: Command has to start with `cp`
```suggestion
        if old_file != new_file or copy == "cp":
```

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        copy, old_file, new_file = command.split()
```
EdAlekseiev, line 3: Firstly, you should check, that `command` contains 3 elements

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3:
+        try:
+            copy, old_file, new_file = command.split()
```
EdAlekseiev, line 4: Don't call `split` twice. You should create variable for storing result of this function and use everywhere you need

EdAlekseiev (changes requested):

vsmutok (approved):
Great!

PR #911, opened 2024-03-31
solution

CharmNiovi:


alonasorochynska (changes requested):
To solve this task it is enough: to check if names of files are same or not, open files, copy content

app/main.py
```diff
@@ -1 +1,34 @@
...
+        raise ValueError(
+            "Expected 2 arguements"
+        )
+
+    if source_parts[0] != source_parts[1]:
+        with (
+            open(source_parts[0], "r") as source,
+            open(source_parts[1], "w") as destination
+        ):
+            destination.write(source.read())
```
YuriiKindrat, line 34: try to avoid using indexes, try to unpack splited command like this:

YuriiKindrat (changes requested):

YuriiKindrat (approved):
GJ

PR #912, opened 2024-03-31
solution

alonasorochynska:


yossosh (approved):
good job!

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_separate = command.split()
+    file_name1 = command_separate[1]
+    file_name2 = command_separate[2]
```
YuriiKindrat, line 4: try to unpack like this 

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if file_name1 != file_name2:
```
YuriiKindrat, line 5: also u should check the command, if command is `cp`

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Well done 🌸

PR #913, opened 2024-03-31
'Solution1'

yossosh:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
```
bwaah1, line 3: Check that the "cp" input command is correct.

bwaah1 (changes requested):

app/main.py
LiudmylaKulinchenko: Fix that. You have this point in your checklist, don't forget to read it before creating a pull request

LiudmylaKulinchenko (changes requested):

YuriiKindrat (approved):
GJ

PR #914, opened 2024-03-31
Implement Copy file task

bwaah1:


Gilgumish (approved):
Looks good but fix flake8 xD

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
```
YuriiKindrat, line 2: u could unpack it like this 
to avoid using indexes

app/main.py
```diff
@@ -1 +1,13 @@
...
+
+    if command_list[0] != "cp":
+        return
+
+    source_file, destination_file = command_list[1], command_list[2]
+
+    if source_file == destination_file:
+        return
```
YuriiKindrat, line 10: no need to do this, try to write backward conditions

YuriiKindrat (changes requested):

app/main.py
YuriiKindrat: also, all tests should pass, fix ur flake8 errors

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, source_file, destination_file = command.split()
+
+    if cp == "cp" or source_file != destination_file:
```
YuriiKindrat, line 4: are u sure about `or` condition here ?

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🌸

PR #918, opened 2024-04-01
Solution

SerhiiMar:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, file_name, copy_name = command.split()
+    if command != "cp":
+        print(f"{command}: command not found")
+        return
+    with open(file_name, "r") as file_in, open(copy_name, "w") as file_out:
+        file_out.write(file_in.read())
```
dkotkod, line 7: In general, everything should work. But I would recommend , before opening the input and output files for copying. It is necessary to check whether the names of the input and output files are different.

fsocie7y (approved):
nice

PR #923, opened 2024-04-01
'Solution'

kssshevchenko:


VladyslavKsh (approved):
Well done

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    parts = command.split()
+
+    if len(parts) == 3 and parts[1] != parts[2]:
+
+        if os.path.exists(parts[1]):
+
+            with (open(parts[1], "r") as file_in,
+                  open(parts[2], "w") as file_out):
+                file_out.write(file_in.read())
```
YuriiKindrat, line 14: try to avoid using indexes - it is really hard to read such code
try to unpack command like this 
var1, var2, var3 = command.split()
Also u should check the command, what if command will not be 'cp' ?

YuriiKindrat (changes requested):

PR #924, opened 2024-04-01
Solution

VladyslavKsh:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    file_name1 = command[1]
+    file_name2 = command[2]
+    if file_name1 == file_name2 and command[0] == "cp":
+        return
+    with open(file_name1, "r") as file_in, open(file_name2, "w") as file_out:
+        for line in file_in.readlines():
```
VitalyBashkiser, line 8: You can do without the readlines() method, since in large projects it will take up a lot of memory; a simple loop for line in file_in is enough for line-by-line search

VitalyBashkiser (approved):
Overall everything looks good

app/main.py
YuriiKindrat: @VitalyBashkiser is right, try to do this in one line without loop

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    file_name1 = command[1]
+    file_name2 = command[2]
+    if file_name1 == file_name2 and command[0] == "cp":
+        return
```
YuriiKindrat, line 6: try to unpack the command like this 

YuriiKindrat (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🌸

PR #925, opened 2024-04-01
Solution#1

VitalyBashkiser:


siaserb (approved):

app/main.py
```diff
@@ -1 +1,39 @@
...
+        print(f"Error: File '{old_file}' not found.")
+    except Exception as e:
+        print(f"An error occurred: {str(e)}")
+
+
+# Example use:
+copy_file("cp file.txt file.txt")  # Does nothing
+
+copy_file("cp file.txt new_file.txt")
+print(open("file.txt").read() == open("new_file.txt").read())  # True
```
YuriiKindrat, line 39: there is no need to leave this checks in ur code )

app/main.py
```diff
@@ -1 +1,39 @@
...
+    # Extract the filenames from the command
+    _, old_file, new_file = parts
+
+    # Check if the old_file and new_file are the same
+    if old_file == new_file:
+        print(
+            "The source and destination filenames are the same. "
+            "Nothing to do."
+        )
+        return
```
YuriiKindrat, line 22: it seems to me, that there will be much less code if u will unpack vars in try except block to validate len of command, and then will write a backward condition to write a file
also u should validate a command)

YuriiKindrat (changes requested):

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    # Validate the command
+    if parts[0] != "cp":
+        print("Invalid command. Please use 'cp' command.")
+        return
```
Oleksl888, line 25: You can check for len of parts, validate cp command and compare file names in one line - and then perform main logic under this if statement so you won't need returns.
Also these comments are self explanatory - no need to clutter code with something that is obvious

app/main.py
```diff
@@ -1 +1,34 @@
...
+
+    try:
+        with open(old_file, "r") as file_in, open(new_file, "w") as file_out:
+            file_out.write(file_in.read())
+        print(f"File '{old_file}' copied to '{new_file}' successfully.")
```
Oleksl888, line 30: You can use `shutil.copy` here.
No need for additional print that file has been copied.
OSError will cover most errors that may occur when handling file

Oleksl888 (changes requested):

Oleksl888 (changes requested):
you haven't fixed previuos comments yet

app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) == 3 and parts[0] == "cp":
```
Oleksl888, line 7: `if len(parts) == 3 and parts[0] == "cp" and old_file != new_file:`
...rest of logic here

Oleksl888 (changes requested):

YuriiKindrat (approved):
GJ

PR #926, opened 2024-04-01
Solution

siaserb:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    function, file_in_name, file_out_name = command.split()
+    if (function == "cp"
```
Oleksl888, line 3: what if there are not exactly 3 elements after command.split()?

app/main.py
```diff
@@ -1 +1,10 @@
...
+            and file_in_name.endswith(".txt")
+            and file_out_name.endswith(".txt")
+            and file_in_name != file_out_name):
+
+        with (open(file_in_name, "r") as file_in,
```
Oleksl888, line 8: How do you handle file-related errors that may occur here?
Also you can use `shutil.copy` here.

Oleksl888 (changes requested):

LiudmylaKulinchenko (approved):
Well done 🔥

PR #927, opened 2024-04-02
'solution'

liudmylaperchevska:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files_name = command.split(" ")
```
pavlopro, line 2: 1.Split by emtry string, not by " "
2.Check that lenght is 3
3.Check that first element is "cp"
4.Unpack it into variables

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if files_name[1] == files_name[2]:
+        return
+
+    with (open(files_name[1], "r") as file_in,
+          open(files_name[2], "w") as files_name[2]):
+        text = file_in.read()
+        files_name[2].write(text)
```
pavlopro, line 9: Pass file_in.read() directly to write method

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #928, opened 2024-04-02
'Solution'

afaidiuk:


app/main.py
```diff
@@ -1 +1,40 @@
-# write your code here
+def copy_file(command: str) -> None:
+    # Split the command into parts
```
Oleksl888, line 2: Remove comments that are self explanatory. If you use split() it is obvious that this is splitting the string

app/main.py
```diff
@@ -1 +1,40 @@
...
+    parts = command.split()
+
+    # Ensure the command has 3 parts
+    if len(parts) != 3:
```
Oleksl888, line 6: all the checks like len of command, file_names etc can be done in one string. Also you  missing check that command = `cp` 
For these three checks no need to print anything - let your program do main logic only if these three conditions are met

app/main.py
```diff
@@ -1 +1,40 @@
...
+        print("Invalid command format. "
+              "Usage: cp <source_file> <destination_file>")
+        return
+
+    # Extract source and destination file names
+    source_file = parts[1]
```
Oleksl888, line 12: this can be done in one line with unpacking

app/main.py
```diff
@@ -1 +1,40 @@
...
+            content = file_in.read()
+
+            # Check if content is not empty
+            if content:
+                # Open the destination file for writing
+                with open(destination_file, "w") as file_out:
+                    # Write the content to the destination file
+                    file_out.write(content)
+                print(f"File '{source_file}' "
+                      f"copied to '{destination_file}' successfully.")
```
Oleksl888, line 34: This can be done in one line with `shutil.copy` or in two lines if you nest open(file_name) keywords.

app/main.py
```diff
@@ -1 +1,40 @@
...
+            else:
+                print(f"File '{source_file}' is empty. No action taken.")
+    except FileNotFoundError:
```
Oleksl888, line 37: OSError will cover all file-related errors here

Oleksl888 (changes requested):

PR #929, opened 2024-04-03
Solution

Ruxxell54:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) == 3 and parts[1] != parts[2]:
+        source_file = parts[1]
+        destination_file = parts[2]
+        print(1)
+        with open(source_file, "r") as file_in,\
```
Oleksl888, line 7: 1. don't use '\' for line breaking
2. You can use `shutil.copy` to simplify all this.

app/main.py
```diff
@@ -1 +1,11 @@
...
+                open(destination_file, "w") as file_out:
+            for line in file_in:
```
Oleksl888, line 9: you can simply use read() on file_in

app/main.py
Oleksl888, line 7: Does your code handle errors that may occur when working with file?

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    if command.startswith("cp "):
+        _, source_file, destination_file = command.split()
```
Oleksl888, line 6: your program still may crash if there is more or less then exactly 3 words in teh command.
THe right design for this would be:
1. Split a command.
2. Check if command has exactly 3 elements, starts with cp and files to copy are not the same
3. rest of code here

Oleksl888 (changes requested):

LiudmylaKulinchenko (commented):
You didn't fix last comment, you should follow the design Ihor discribed:

1. Split a command.
2. Check if command has exactly 3 elements, starts with cp and files to copy are not the same
3. rest of code here

LiudmylaKulinchenko (changes requested):
You didn't fix last comment, you should follow the design Ihor discribed:

1. Split a command.
2. Check if command has exactly 3 elements, starts with cp and files to copy are not the same
3. rest of code here

PR #930, opened 2024-04-03
Solution for py-copy-file

NinqueRaven13:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    orig_file, new_file = command[3:].split(" ")
```
vsmutok, line 2: u could unpack it like this
to avoid using indexes

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if orig_file == new_file:
+        return
```
vsmutok, line 4: better to raise some Exception, so that the user will understand if something goes wrong

app/main.py
```diff
@@ -1 +1,7 @@
...
+    with open(orig_file, "r") as f1, open(new_file, "w") as f2:
+        for line in f1.readlines():
+            f2.write(line)
```
```suggestion
        f2.write(f1.read())
```

vsmutok: also `f1, f2` not clear variable name

vsmutok (changes requested):
image

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, orig_file, new_file = command.split()
```
vasylhnatiuk, line 2: you have to check if command has 3 arguments

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #931, opened 2024-04-03
solution

Githi54:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command) < 3:
+        return
+    operation, file_name_to_copy, new_file_name = command.split(" ")
```
LiudmylaKulinchenko, line 4: no need to use space inside split method, it uses space by default

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good! 🌼

PR #934, opened 2024-04-05
Solution

onyevyerov:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp_command, file_for_copy, new_file = command.split()
```
EdAlekseiev, line 2: Firstly, you should check, that command contains 3 elements and start with `cp`

EdAlekseiev (changes requested):

ArsenPidhoretskyi (approved):
Well done!

PR #937, opened 2024-04-07
solved

serhii-chaban:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts_of_command = command.split()
+    file_source = parts_of_command[1]
+    file_new = parts_of_command[2]
+    condition = [
+        command[:2] == "cp",
```
Dimosphen1, line 6: Use unpacking after checking the length of the element after split to avoid indexing, make code more descriptive and reducing the number of conditions to check

app/main.py
```diff
@@ -1 +1,15 @@
...
+        file_source != file_new,
+        len(parts_of_command) == 3,
+        ".txt" in file_source,
+        ".txt" in file_new,
```
Dimosphen1, line 10: You could copy not only files with `.txt` extension, remove this check

app/main.py
```diff
@@ -1 +1,15 @@
...
+    ]
+    if all(condition):
```
Dimosphen1, line 12: You could place checks here without `condition` variable allocation

app/main.py
```diff
@@ -1 +1,15 @@
...
+        with open(file_source, "r") as source, open(file_new, "w") as copied:
+            for line in source.readlines():
+                copied.write(line)
```
Dimosphen1, line 15: Use `read()` operation on file that you will pass directly to the `write()` to avoid iteration

Dimosphen1 (changes requested):

PR #940, opened 2024-04-08
solution

haldaniko:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if (len(command.split()) == 3
+            and command.split()[0] == "cp"
+            and command.split()[1] != command.split()[2]):
+        file_in, file_out = command.split()[1], command.split()[2]
```
vsmutok, line 5: DRY

you should use `command.split()` only once

vsmutok (changes requested):

PR #942, opened 2024-04-09
Solution

ggeorgiyy:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    spl_com = command.split(" ")
```
vasylhnatiuk, line 5: spl_com is a bad name

app/main.py
```diff
@@ -1 +1,11 @@
...
+    with open(spl_com[1], "r") as file_in, open(spl_com[2], "w") as file_out:
```
vasylhnatiuk, line 6: you should check if command has 3 arguments to avoid error here

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
.

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    path = command.split(" ")
```
```suggestion
    path = command.split()
```
Also `path` not clear name

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if len(path) == 3:
+        with open(path[1], "r") as file_in, open(path[2], "w") as file_out:
+            if file_in == file_out:
+                pass
```
vsmutok, line 9: why `pass`?
the user will not understand what went wrong, it is better to print something

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    path = command.split()
+    if len(path) == 3:
+        with open(path[1], "r") as file_in, open(path[2], "w") as file_out:
+            if file_in == file_out:
+                print("Your command don't allow to copy file")
```
EdAlekseiev, line 9: You should do it before open files

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #943, opened 2024-04-09
Solution

frontbastard:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, file1, file2 = command.split(" ")
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #944, opened 2024-04-09
solution

nrjss67:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    old_file, new_file = command.split()[1:]
```
vasylhnatiuk, line 2: You should check if command has 3 arguments before it

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3:
+        comm, old_file, new_file = command.split()
```
EdAlekseiev, line 3: Don't call split twice. You should create variable for storing result of split and using where you need

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #945, opened 2024-04-10
Solution

fishray33:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd_mapped = command.split()
+    if len(cmd_mapped) != 3:
+        return
+    if cmd_mapped[1] == cmd_mapped[2]:
+        return
+    if cmd_mapped[0] != "cp":
+        return
```
NazarSheremet, line 8: All these checks can be combined

NazarSheremet (changes requested):
Minor changes

app/main.py
fishray33: Done!

YuriiSmolii (approved):
LGTM

PR #946, opened 2024-04-11
copy_function_v1.0

YegorVolkov:
Наче все працює

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    """
+    this function designed for copying two files
+    :param command: "cp file_to_copy.format new_file.format"
+    """
+    file_1 = command.split(" ")[1]
+    file_2 = command.split(" ")[2]
+    with open(file_1, "r") as file_in, open(file_2, "w") as file_out:
+        file_out.write(file_in.read())


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    """
+    this function designed for copying two files
+    :param command: "cp file_to_copy.format new_file.format"
+    """
+    file_1 = command.split(" ")[1]
+    file_2 = command.split(" ")[2]
```
NazarSheremet, line 7: It can be done simpler
```suggestion
    cmd, file_1, file_2 = command.split()

```

NazarSheremet (changes requested):
Also need to implement 2 additional checks
image
image

app/main.py
YegorVolkov: Wow! 
very interesting way!
Thank you!

YegorVolkov:
1) I did it using Exeptions in order to practice and to receive a review if I used it properly
2) I could not find the way to delete a commit from a remote repo (of course I would not do it if it is public)
if it was public I would use `git revert -> git push`
but if it is my personal remote repo & I wish to make the log clear -> How can I manipulate remote repo in a way like:
`git reset Hash ` on local repository
I could delete the branch from remote repo using `git push origin :dev` but in this way, I think I would loose all previous comments to commits... Not mentioning tht I would need to refork it again to make MateAcademy recognise my work

app/main.py
NazarSheremet: It's not the functionality of a `split()` function. It only returns your list and after you assign the value of the list to variables.
Something like this 
```
list = command.split() # the value list will be [cmd, file1, file2]
cmd, file_1, file_2 = list
```
 So it's more about work with list

NazarSheremet (approved):
Appvoved, but looks like overhead solution. It was only necessary to add two conditions: one to return nothing if the file is the same, and another to raise an exception or output that the command is incorrect.

PR #947, opened 2024-04-11
Done

MikolaRozhansky:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        cmd = command.split()
+        command_name = cmd[0]
+        source_name = cmd[1]
+        copy_name = cmd[2]
```
EdAlekseiev, line 6: You should use unpacking
```suggestion
        command_name, source_name, copy_name = cmd
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+    except Exception:
+        print("Error in command argument!")
+        return None
```
EdAlekseiev, line 9:
```suggestion
        return
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+    if command_name == "cp":
+        if source_name != copy_name:
```
EdAlekseiev, line 12: You can join two if to one
```suggestion
    if command_name == "cp" and source_name != copy_name:
```

EdAlekseiev (changes requested):

vsmutok (approved):
Great!

PR #951, opened 2024-04-14
'Solution'

Sonemon:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, file1, file2 = command.split(" ")
```
vasylhnatiuk, line 2: you should check if command has 3 arguments

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.count(" ") != 2:
```
```suggestion
    if command.count(" ") != 3:
```

vasylhnatiuk: also check if command is `cp`

vasylhnatiuk (changes requested):

app/main.py
Sonemon: Why do we check for 3 spaces and not 2? In the command "cp file1 file2" - 2 spaces?

Sonemon: > 
Why do we check for 3 spaces and not 2? In the command "cp file1 file2" - 2 spaces?

vasylhnatiuk: sorry, better use len() here isnstead of count
if len(command.split()) != 3:

vasylhnatiuk (changes requested):

app/main.py
Sonemon: Okay, I got it. Thank you.

YukihiroSM (approved):
Looks good to me! Nicely done!

PR #954, opened 2024-04-17
copy file solution

TenbReala:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copi_file(command: str) -> None:
+    command_parts = command.split()
+    if command_parts[0] == "cp" and command_parts[1] != command_parts[2]:
```
EdAlekseiev, line 3: Firstly, you should check that `command_parts` contains 3 elements
```suggestion
    if len(command_parts) == 3 and command_parts[0] == "cp" and command_parts[1] != command_parts[2]:
```

EdAlekseiev (changes requested):

PR #964, opened 2024-05-01
Solution

sind14:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    f_list = command.split()
+    if f_list[1] == f_list[2]:
```

vasylhnatiuk (changes requested):

EdAlekseiev (approved):
Good job

PR #965, opened 2024-05-01
solution

NAZ-KOM:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_input = command.split()
+    if splitted_input[1] != splitted_input[2]:
```
vsmutok, line 3: You should check that command have exactly 3 parameters, command is exactly "cp"

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_input = command.split()
+    if (len(splitted_input) == 3
+            and splitted_input[0] == "cp"
+            and splitted_input[1] != splitted_input[2]):
```
Dimosphen1, line 5: It would be better to use unpacking instead of retrieving values by index in the future in such cases to improve code readability.

Dimosphen1 (approved):
LGTM, one comment is left for consideration

PR #966, opened 2024-05-02
'Solution'

v-shutov:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    split_command = [word for word in command.split(" ")]
+    command = split_command[0]
+    file_name_in = split_command[1]
+    file_name_out = split_command[2]
```
NazarSheremet, line 6: It can be done simpler

```suggestion
    command, file_name_in, file_name_out = command.split()
```

NazarSheremet (changes requested):

app/main.py
v-shutov: thx, changed

YuriiSmolii (approved):
LGTM

PR #970, opened 2024-05-04
triukhan solution

triukhan:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_command = command.split(" ")
```
LiudmylaKulinchenko, line 2: no need to use space inside split method, it uses space by default

app/main.py
```diff
@@ -1 +1,11 @@
...
+    if splitted_command[0] != "cp" or len(splitted_command) != 3:
+        print("Incorrect command")
+        return
+    if splitted_command[1] == splitted_command[2]:
+        return
```
LiudmylaKulinchenko, line 7: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,11 @@
...
+    with (open(splitted_command[1], "r") as file_in,
+          open(splitted_command[2], "w") as file_out):
+        text = file_in.read()
+        file_out.write(text)
```
LiudmylaKulinchenko, line 11: It's ok to avoid creating additional variable here

LiudmylaKulinchenko (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, old_file, new_file = command.split()
+    if command != "cp" or len(command.split()) != 3:
```
LiudmylaKulinchenko, line 3: You should check command length before splitting it, because it will trigger an error during splitting

LiudmylaKulinchenko (changes requested):

app/main.py
triukhan: згоден крінж, переробив

LiudmylaKulinchenko (approved):
Well done 💪💪

PR #971, opened 2024-05-06
Done

Stekloduv:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd, source_file, dest_file, *_ = command.split()
```
pavlopro, line 2: Before unpacking, it would be good to check that the length of the split result equals 3

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if cmd != "cp" or source_file == dest_file:
+        return
+    try:
```
pavlopro, line 5: Please remove this try - expect condition

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #973, opened 2024-05-07
completed

aniashev:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    new_list = command.split(" ")
+    old_name = new_list[1]
```
vasylhnatiuk, line 3: you should check if new_list has 3 arguments before it

vasylhnatiuk: and you can use `old_name, new_name  = new_list[1:]`

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #976, opened 2024-05-12
Solution

AndrewChugreev:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
```
Dimosphen1, line 2: There is no need to pass `" "` to split, it is a default option, remove it
```suggestion
    command_list = command.split()
```

Dimosphen1 (changes requested):

vasylhnatiuk (approved):

PR #978, opened 2024-05-14
Solution

Jejemira:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(file_name: str) -> None:
+    words_of_file_name = file_name.split(" ")
```
```suggestion
    words_of_file_name = file_name.split()
```

app/main.py
```diff
@@ -1 +1,9 @@
...
+    old_file_name = words_of_file_name[1]
+    new_file_name = words_of_file_name[2]
```
EdAlekseiev, line 4: You should check that `words_of_file_name` contains 3 elements before unpacking value to avoid `IndexError`

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #981, opened 2024-05-19
Hell yeah solution

RuslanLevchuk:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+import shutil
```
vsmutok, line 1: don't use `shutil` write your custom logic

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+
+def copy_file(command: str) -> None:
+    command = command.split(" ")
```
```suggestion
    command = command.split()
```

vsmutok (changes requested):

app/main.py
RuslanLevchuk: okay... I knew it..but...

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if len(command) == 3 and command[0] == "cp":
+        with (
+            open(command[1], "r") as read_file,
+            open(command[2], "w") as write_file
+        ):
+            for line in read_file.readlines():
+                write_file.write(line)
+        return
+    print("Invalid command")
+    return
```
vsmutok, line 12:
```suggestion
    print("Invalid command")
```

`return` is redundant here, since the function will end anyway

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) == 3 and command[0] == "cp":
+        with (
+            open(command[1], "r") as read_file,
+            open(command[2], "w") as write_file
+        ):
+            for line in read_file.readlines():
+                write_file.write(line)
```
vsmutok, line 9:
```suggestion
            write_file.write(read_file.read())
```
no need to use loop here

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #982, opened 2024-05-20
Solution

taras-tkachuk:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    old_file_name = command.split(" ")[1]
+    new_file_name = command.split(" ")[2]
```
EdAlekseiev, line 3: Don't use `command.split` twice. You should create a separate variable for storing result of `command.split` and use it where you need

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #983, opened 2024-05-21
Solution

Anon920:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split(" ")
+    command_old = command_split[1]
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,4 +1,6 @@
 def copy_file(command: str) -> None:
+    if len(command) > 4:
```
vasylhnatiuk, line 2: but what if command has less then 3 arguments?

app/main.py
```diff
@@ -1,4 +1,6 @@
...
+        exit()
```
vasylhnatiuk, line 3: print message for user that they should put 3 arguments

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command) > 4:
```
```suggestion
    if len(command) == 3:
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+        print("Error: You need to provide exactly 2 arguments")
+    command_split = command.split(" ")
+    command_old = command_split[1]
```
```suggestion
    _, command_old, command_new  = command_split
```

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: use more accurate names, and use DRY

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+import os
+
+
+def copy_file(command: str) -> None:
+    command_split = command.split()
+    if len(command_split) != 3 or command[0] != "copy":
+        return
```
vsmutok, line 7: How should a user understand if they enter an incorrect command?
Please add `print` with some informative  message before `return`

app/main.py
```diff
@@ -1 +1,19 @@
...
+    _, old_file_name, new_file_name = command_split
+
+    if old_file_name == new_file_name:
+        return
```
vsmutok, line 11: the same

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #984, opened 2024-05-21
Solution

Shayba228:


vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+import os
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        return
```
vasylhnatiuk, line 7: plase write some message for use that his command is not correct

vasylhnatiuk (changes requested):

PR #988, opened 2024-05-26
solution

AndrewwSSS:


app/aboba.txt
Oleg-s-s, line 1: ?

app/test.txt
Oleg-s-s, line 1: ?

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    args = command.split(" ")
+    if args[1] == args[2] or args[0] != "cp":
+        return
```
Oleg-s-s, line 4: You can split in multiple vars and then check.
Better reading + optimization.

Oleg-s-s (changes requested):

app/aboba.txt
AndrewwSSS: test case

app/test.txt
AndrewwSSS: test case

Oleg-s-s (approved):
👍

PR #991, opened 2024-05-31
Solution

TsvetanKichuk:


app/copy_file.py
vsmutok, line 1: remove this file from PR

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,26 @@
...
+
+    if not os.path.isfile(source_file):
+        print(f"Source file '{source_file}' not found.")
+        return
+
+    with (open(source_file, "r") as file_in,
+          open(destination_file, "w") as file_out):
+
+        data = file_in.read()
+        file_out.write(data)
```
Dimosphen1, line 24: There is no need to create `data` variable, you could pass `file_in.read()` directly to `file_out.write`

app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+import os
+
+
+def copy_file(my_file: str) -> None:
+    parts = my_file.split()
```
Dimosphen1, line 5: You could follow another approach by avoiding value retrieval by indexing. In this way code would be more descriptive 
```suggestion
    command, source_file, destination_file, *_ = my_file.split()
```
Follow this approach in the future

Dimosphen1 (approved):
LGTM, several comments were left for consideration

PR #996, opened 2024-06-02
'Solution'

BohdanKuzik:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None | bool:
+    command = command.split(" ")[1:]
+    if command[0] != command[1]:
```
DevSpaciX, line 3: It is better to unpuck your commend[0] and command[1] into a viriables fot better understanding . You can do it in one line

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None | bool:
+    command = command.split(" ")[1:]
```
DevSpaciX, line 2: Also split method by default splitting by " "

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if command[0] != command[1]:
+        with (open(command[0], "r") as file_in,
+              open(command[1], "w") as file_out):
+            file_out.write(file_in.read())
+        if open("file.txt").read() == open("new_file.txt").read():
+            return True
+    return
```
DevSpaciX, line 9: You can `call retrun` at the start of your func , and just check if file_name 1 == file_name 2 and then just write the file into another

app/main.py
```diff
@@ -1 +1,10 @@
...
+def copy_file(command: str) -> None | bool:
+    command = command.split()[1:]
+    input_file, output_file = command[0], command[1]
+    if input_file != output_file:
+        with (open(input_file, "r") as file_in,
+              open(output_file, "w") as file_out):
+            file_out.write(file_in.read())
+        if open("file.txt").read() == open("new_file.txt").read():
+            return True
+    return
```
Oleg-s-s, line 10: Do you need this code?
Looks like you have no belief in python built ins)))

Oleg-s-s (changes requested):

PR #998, opened 2024-06-03
Solution

Yuruch:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, first_file, second_file = command.split(" ")
```
DevSpaciX, line 2: `Split()` method by default splitting by `" " value `, so you can just write `split()`

DevSpaciX (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if first_file == second_file or command != "cp":
+        return
+    with open(first_file, "r") as file_in, open(second_file, "w") as file_out:
+        for line in file_in:
+            file_out.write(line)
```
DevSpaciX, line 7: Also you do not need `for loop` here

PR #1000, opened 2024-06-03
Solution

voboian:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    old_file, new_file = parts[1], parts[2]
```
Dimosphen1, line 3: Perform unpacking here instead of retrieving values by index

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if old_file != new_file:
```
Dimosphen1, line 4: Make sure additionally that the command is `cp` before processing further

Dimosphen1 (changes requested):

PR #1002, opened 2024-06-04
add solution

jpdreamthug:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands_list = command.split()
+    if commands_list[0] == "cp":
+        if commands_list[1] != commands_list[2]:
```
DevSpaciX, line 4: Do no create additional nesting . You can check this in one line using operator `and`. Also it is better to unpuck the value witch you calling by indexes into the viriables , and reuse them

DevSpaciX (changes requested):

DevSpaciX (approved):
ultra good job!

PR #1005, opened 2024-06-04
solution

NikitaNikolaitsev:


DevSpaciX (approved):

DevSpaciX (commented):
But it is better to unpuck the values into the viriables for future : )

PR #1006, opened 2024-06-04
Solution V.1

SemenetsSergii:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    com_from = command[1]
+    com_to = command[2]
+    if com_from[1] == com_to[2]:
+        pass
+    with open(com_from, "r") as f_in, open(com_to, "a") as f_out:
```
DevSpaciX, line 7: Think more about naming viriables . `f_in`  `com_to`  and else are not informative

DevSpaciX (changes requested):

DevSpaciX (approved):
file_in , file_out - better viriables imho : )
`com_to` ? May be come_to ? Anyway next time think more about it : )

PR #1008, opened 2024-06-04
'Solution'

Oleh323:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    com_from = command[1]
+    com_to = command[2]
+    if com_from[1] == com_to[2]:
```
DevSpaciX, line 5: Why do you using indexes on str ? com_from it is already spited value . Also think more about viriables naming . `com_to , com_from` it is not very good naming inho

app/main.py
```diff
@@ -1 +1,8 @@
...
+        pass
+    with open(com_from, "r") as in_, open(com_to, "a") as out_:
```
DevSpaciX, line 7: as `in_` ? Same it can be named `input_file` or smth

DevSpaciX (changes requested):

DevSpaciX (approved):

PR #1009, opened 2024-06-04
sol1

Unlie9:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
+    copy_file = command[1]
+    new_file = command[2]
+    if copy_file == new_file:
+        new_file = "new_" + command[2]
```
DevSpaciX, line 6: You can call `return` here if same file already exists

app/main.py
```diff
@@ -1 +1,16 @@
...
+
+    content = []
+    with open(copy_file) as f:
+        lines = f.readlines()
+
+        for line in lines:
+            content.append(line)
+
+    with open(new_file, "w") as f:
+        f.writelines(content)
```
DevSpaciX, line 16: You can do it in 2 line . On the first open 2 files . 
On the second call `write()` method and inside that `write()` you can call `read()` method for your second file

DevSpaciX (changes requested):

DevSpaciX (approved):
Mega nice : )

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
```
DevSpaciX, line 2: `split already takes " " by default` param , so you can call just `split() `

PR #1013, opened 2024-06-04
Done Task!

TkachukTara:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split()
+    if commands[1] != commands[2] and commands[0] == "cp":
```
DevSpaciX, line 3: Put your `commands[1] `... into a viriables , cuz it is not clear for me that it is file names here : )

DevSpaciX (changes requested):

PR #1021, opened 2024-06-05
'Solution'

Irina17191:


app/main.py
```diff
@@ -1 +1,25 @@
...
+    if original_file == copied_file:
+        return None
+
+    # we can check if file with same name already exists
+    try:
+        with open(copied_file, "r"):
+            pass
+        return None
+    except FileNotFoundError:
+        pass
```
DevSpaciX, line 18: You can delete this part of code. That check is not necessery for now

app/main.py
```diff
@@ -1 +1,25 @@
...
+
+    with (open(original_file, "r") as file_in,
+          open(copied_file, "w") as file_out):
+        for line in file_in:
+            file_out.write(line)
```
DevSpaciX, line 23: you can call method `read() on the file_in`  inside method `write() on the file_out` . That way you could escape of using loop

app/main.py
```diff
@@ -1 +1,25 @@
...
+from typing import Optional
+
+
+def copy_file(command: str) -> Optional[str]:
+
+    original_file = command.split(" ")[1]
+    copied_file = command.split(" ")[2]
+
+    if original_file == copied_file:
+        return None
```
DevSpaciX, line 10: Also if you call just `return` is will return `None` too : ) So that None a bit redundat imho

DevSpaciX (changes requested):

PR #1023, opened 2024-06-08
Solution

MaksymProtsak:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd_list = command.split()
+    if cmd_list[1] != cmd_list[2]:
+        with (open(cmd_list[1], "r") as old_f,
+              open(cmd_list[2], "w") as new_f):
+            new_f.write(old_f.read())
```
vsmutok, line 6: * Make sure you check that the `cp` input command is correct.
* Make sure that the length of `cmd_list` is three

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd_list = command.split()
+
+    if (
+        (cmd_list[0] == "cp")
+        and (len(cmd_list) == 3)
```
vasylhnatiuk, line 6: this condition should be first

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd_list = command.split()
+
+    if (
```
vasylhnatiuk, line 4: add `else` with message that user should write correct data

vasylhnatiuk (changes requested):

app/main.py
MaksymProtsak: Done.

app/main.py
MaksymProtsak: Done.

app/main.py
MaksymProtsak: Done.

vasylhnatiuk (approved):

MaksymProtsak:
🕺

PR #1025, opened 2024-06-10
Solution

Artemq77:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    source_file, dest_file = parts[1], parts[2]
```
Dimosphen1, line 3: Unpack the parts without indexing and preserving the command name
```suggestion
    command_name, source_file, dest_file, *_ = parts
```

app/main.py
```diff
@@ -1 +1,8 @@
...
+
+    if source_file != dest_file:
```
Dimosphen1, line 5: Make sure that the command is actually `cp` as well

Dimosphen1 (changes requested):

PR #1028, opened 2024-06-14
"solution"

arsenmarkotskyi:


arsenmarkotskyi:
i understand that its is not necessary :
            print(f"all info from {old_file} was copy to {new_file}")


copy_file("cp file.txt new_file.txt")

but we don’t have repo with tests here, so left it instead of tests))

Polyakiv-Andrey (approved):
Good job!

PR #1032, opened 2024-06-16
done

Alex93Troy:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    first_file, second_file = command[3:].split(" ")
```
vsmutok, line 2: before that you should check that it won't cause an error
* check that the  `command` length is equal to 3
* check that the first element is `cp`

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if first_file == second_file:
+        pass
```
vsmutok, line 4: better to `print` some informative message

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if len(parts) != 3 or parts[0] != "cp" or parts[1] == parts[2]:
+        print("Data isn't correct,"
+              " please enter other"
+              " name command or other names files")
+    else:
+        content = []
+        with open(parts[1], "r") as file_data:
+            content = file_data.read()
+        with open(parts[2], "w") as file_copy:
+            file_copy.write(content)
```
vsmutok, line 12: use only one `with `

vsmutok (changes requested):

EdAlekseiev (approved):

PR #1034, opened 2024-06-17
Solution

danilsiv:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command.split(" ")
```
```suggestion
    file_names = command.split()
```

app/main.py
```diff
@@ -1 +1,7 @@
...
+    if file_names[1] == file_names[2]:
```
vasylhnatiuk, line 3: you should check if command is correct and if command has 3 arguments before it

app/main.py
```diff
@@ -1 +1,7 @@
...
+        return None
```
vasylhnatiuk, line 4: write message for user instead of none

vasylhnatiuk (changes requested):

PR #1039, opened 2024-06-24
solution1

KovbasinskyV:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    action, file, new_file = command.split(" ")
```
vasylhnatiuk, line 2: you should check if command has 3 arguments before it and the first argument is correct

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splited_command = command.split(" ")
+    action, file, new_file = splited_command
+
+    if action == "cp" and len(splited_command) == 3:
```
vasylhnatiuk, line 5: you should check  `len(splited_command) == 3` before `3` line

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split(" ")
```
pavlopro, line 2: `split` method uses space by default so you can simplify to `command.split()`

pavlopro (changes requested):

PR #1040, opened 2024-06-24
Solution

EugeneDolinskiy:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_command = command.split("")
```
Dimosphen1, line 2: There is no need to pass an argument here, since you should split by space and it is a default option

app/main.py
```diff
@@ -1 +1,8 @@
...
+    with (open(splitted_command[1], "r") as f_original,
+          open(splitted_command[2], "w") as f_copy):
+        if splitted_command[1] == splitted_command[2]:
+            pass
```
Dimosphen1, line 6: Perform this check earlier before entering a context manager

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_command = command.split("")
+    with (open(splitted_command[1], "r") as f_original,
+          open(splitted_command[2], "w") as f_copy):
+        if splitted_command[1] == splitted_command[2]:
```
Dimosphen1, line 5: Make sure that your command is actually `cp` as well

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_command = command.split("")
+    with (open(splitted_command[1], "r") as f_original,
+          open(splitted_command[2], "w") as f_copy):
```
Dimosphen1, line 4: Avoid indexing, use unpacking instead to make code more descriptive

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    splitted_command = command.split()
+    original_file = splitted_command[1]
```
pavlopro, line 3: You should check the length of split command before retrieve values
It should be exactly 3

app/main.py
```diff
@@ -1 +1,15 @@
...
+    copy = splitted_command[2]
+    cp_command_check = splitted_command[0]
+
+    if not cp_command_check == "cp":
```
pavlopro, line 7: Please combine the two conditions into one 'if'

pavlopro (changes requested):

Polyakiv-Andrey (approved):
Good job!

PR #1043, opened 2024-06-25
Solution

silchenko-arsen:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(commands_str: str) -> None:
+    commands_list = commands_str.split()
+    file_out = commands_list[1]
+    file_in = commands_list[2]
```
NazarSheremet, line 4: In this way looks better
```suggestion
    commands_list, file_out, file_in = commands_str.split()
```

NazarSheremet (changes requested):
Fix suggestions and need to add checks if command equal `cp` and have 3 parts

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(commands_str: str) -> None:
+    commands_list, file_out, file_in = commands_str.split()
+    if (len(commands_list) == 3 and commands_list[0] == "cp"
```
NazarSheremet, line 3: This won't work. The variable `commands_list `will be a `str `and equal to `cp`, so the condition will be false.

NazarSheremet (changes requested):

NazarSheremet (approved):
Good job!

PR #1045, opened 2024-06-26
Try Solution

D0mln0:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cmd_ls = command.split(" ")
```
```suggestion
    cmd_ls = command.split()
```

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if cmd_ls[1] == cmd_ls[2]:
```
vasylhnatiuk, line 3: you check if command has 3 arguments before it, and raise error with message if not

vasylhnatiuk: also check if command is correct

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk: do it as well and `cmd_ls` is not clear name, fix it

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #1046, opened 2024-06-26
'Solution'

hhorhe:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3 or parts[0] != "cp":
+        return
```
```suggestion
    if len(parts) != 3 or parts[0] != "cp":
        print("some informative message")
        return
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    source_file = parts[1]
+    dest_file = parts[2]
+
+    if source_file == dest_file:
+        return
```
vsmutok, line 11:
```suggestion
    if source_file == dest_file:
        print("some informative message")
        return
```

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3 or parts[0] != "cp":
+        return
+
+    source_file = parts[1]
+    dest_file = parts[2]
```
```suggestion
    _, source_file, dest_file = parts
```
better use unpacking in this case

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #1047, opened 2024-06-27
solution

midetc:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    command, file_name_input, file_name_output = command_list
```
EdAlekseiev, line 3: Firstly, you should check that command_list contains 3 elements to avoid errors

EdAlekseiev (changes requested):

vasylhnatiuk (approved):

PR #1049, opened 2024-06-27
Solution

PanValeriy:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
+    file_name_in = command_list[1]
```
vasylhnatiuk, line 3: you should check if comannd has 3 arguments before it and if command is correct

vasylhnatiuk: and write message if command is not correct

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
+    if len(command_list) != 3:
+        raise RuntimeError("Command is not correct!")
+    if command_list[0] != "cp":
+        raise NameError("Expected 'cp' command")
+
+    file_name_in = command_list[1]
```
vasylhnatiuk, line 8:
```suggestion
    file_name_in, file_name_out = command_list[1:], 
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1050, opened 2024-06-28
add solution

TamaraBuchak:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_name = command.split(" ")
```
```suggestion
    file_name = command.split()
```

app/main.py
```diff
@@ -1 +1,20 @@
...
+
+    if (
+        file_name[0] != "cp"
+        or len(file_name) != 3
+        or file_name[1] == file_name[2]
+    ):
+        return
```
vasylhnatiuk, line 9: you should write message that command is wrong

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1051, opened 2024-06-28
Solution

zave52:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file1 = command.split(" ")[1]
+    file2 = command.split(" ")[2]
```
NazarSheremet, line 3: This way looks better 
```suggestion
    command, file_out, file_in = commands_str.split()
```

NazarSheremet (changes requested):
Need to fix suggestions

YuriiSmolii (approved):
Well done!

PR #1053, opened 2024-07-05
solution

Anatolik-ua:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None | str:
+    command_name, file_in, file_out = command.split()
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1054, opened 2024-07-05
'Solution'

dryzhenko:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+    source_file = parts[1]
```
vasylhnatiuk, line 6: you should check if command has 3 argumnts before it

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) == 3:
+        source_file = parts[1]
```
vasylhnatiuk, line 8:
```suggestion
        source_file, destination_file = parts[1:]
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+        destination_file = parts[2]
```
vasylhnatiuk, line 9: and delete this

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) == 3:
```
vasylhnatiuk, line 7: you should write a message for user to show that command is not correct

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1055, opened 2024-07-06
Dev

VasylRysan:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
+        return
```
vsmutok, line 3: how will the user understand that something went wrong? 
give an informative message

app/main.py
```diff
@@ -1 +1,12 @@
...
+    operation, current_file, new_file = command.split()
+    if current_file == new_file:
+        return
```
vsmutok, line 6: the same

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
+        return
+    operation, current_file, new_file = command.split()
```
vsmutok, line 4: `split()` - don't use it twice, use it only once

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1058, opened 2024-07-10
Solution

tylerj231:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+import shutil
+
+source = "/Users/jacktyler/py-copy-file/app/file.txt"
+destination = "/Users/jacktyler/py-copy-file/app/new_file.txt"
```
YukihiroSM, line 4: You shouldn't put your local addresses into the code. It should be working with any address.

app/main.py
```diff
@@ -1 +1,16 @@
...
+def copy_file(source: str, destination: str) -> None:
+    try:
+        shutil.copy(source, destination)
+    except shutil.SameFileError:
+        print("Cannot copy the same file!")
+    else:
+        print("File has been copied successfully")
+
+
+print(copy_file(source, destination))
```
YukihiroSM, line 16: remove this line as well, please.

YukihiroSM (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
...
+import shutil
+
+
+def copy_file(source: str, destination: str) -> None:
+    try:
+        shutil.copy(source, destination)
+    except shutil.SameFileError:
+        print("Cannot copy the same file!")
+    else:
+        print("File has been copied successfully")
```
vsmutok, line 10: please don't use `shutil` here 
write your own logic

vsmutok (changes requested):

app/main.py
tylerj231: Okay

vasylhnatiuk (approved):

PR #1060, opened 2024-07-11
Add solution

MRozdoba:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_line, original_file, new_file = command.split(" ")
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from app.errors import CommandLineError
+
+
+def copy_file(command: str) -> None:
+
+    if len(command.split(" ")) != 3 or command.split(" ")[0] != "cp":
```
Dimosphen1, line 6: There is no need to pass space as split delimiter, it is a default option

app/main.py
```diff
@@ -1 +1,15 @@
...
+        raise CommandLineError("Command line is incorrect.")
+
+    command_line, original_file, new_file = command.split(" ")
```
Dimosphen1, line 9: Do not perform split multiple times, define a variable, write split to it, and after that you can assess length and unpack this variable

Dimosphen1 (changes requested):

pavlopro (approved):
Good Job!

PR #1063, opened 2024-07-16
"Solution py-copy-file"

Olga-Gynguliak:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split(" ")
+
+    if len(parts) < 3 or parts[0] != "cp":
+        return
```
vasylhnatiuk, line 5: write message for user that command is wrong

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split(" ")
```
```suggestion
    parts = command.split()
```

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split("Command is wrong!")
```
vasylhnatiuk, line 2: why did u do that? it doesn't make any sense. pls read how `split` works and add  `print("Command is wrong!") `after 4 line

vasylhnatiuk (changes requested):

EdAlekseiev (approved):

PR #1064, opened 2024-07-16
created copy file realisation

MasakDirt:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command.split(" ")
```

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if file_names[0] != "cp" or file_names[1] == file_names[2]:
+        return
```

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import shlex
+
+
+def copy_file(command: str) -> None:
+    commands = shlex.split(command)
+    source_file = commands[1]
+    destination_file = commands[2]
```
oliasymak24, line 7: commands, source_file,  destination_file = command.split() will be better to use this way

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1066, opened 2024-07-17
Resolved Copy file task

viktor-77:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, copy_from, copy_to = command.split()
```
Dimosphen1, line 2: Do not omit the command name, you should add a check that the command is actually `cp` before performing any other actions

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
```
pavlopro, line 1: Would be good to perform split and save result into variable. After check that lenght of this is exactly 3. Then you can do unpacking

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1067, opened 2024-07-18
'Solution'

Carn1vor3:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+from typing import Any
+
+
+def copy_file(command: str) -> Any:
+    cp, old_file, new_file = command.split()
```
vasylhnatiuk, line 5: you should check if command has 3 arguments and cp is correct command

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+from typing import Any
+
+
+def copy_file(command: str) -> Any:
+    if len(command.split()) != 3 or command.split()[0] != "cp":
+        return
```
vasylhnatiuk, line 6: write message for user

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
...
+from typing import Any
+
+
+def copy_file(command: str) -> Any:
+    if len(command.split()) != 3 or command.split()[0] != "cp":
+        print("Invalid command format."
+              " Use: cp <source_file> <destination_file>")
+        return
+
+    cp, old_file, new_file = command.split()
```
vsmutok, line 10: now you use `command.split()` three times, try to optimize and leave only one use `command.split()`

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1068, opened 2024-07-19
Implement copy_file function

anton-bilous:


app/main.py
```diff
@@ -1 +1,22 @@
...
+def copy_file(command: str) -> None:
+    try:
+        command, source_name, destination_name = shlex.split(command)
+        if command != "cp":
+            raise ValueError("The command is not 'cp'")
+    except ValueError as e:
+        raise ValueError("The command is not valid") from e
+
+    if source_name == destination_name:
+        return
```
oliasymak24, line 13: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    with open(source_name, "rb") as source, open(
+        destination_name, "wb"
+    ) as destination:
+        destination.write(source.read())
+
+
+if __name__ == "__main__":
+    copy_file("cp README.md 'Some file.md'")
```
oliasymak24, line 22: Don't use test call of function here

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+import shlex
```
oliasymak24, line 1: I think no need here to use this library
It is useful if names of file have blanks, but in your test case you can just use simple split

oliasymak24 (changes requested):

oliasymak24 (approved):
Good job!

PR #1070, opened 2024-07-19
Solution

Morphin20th:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file, new_file = command.split()[1:3]
```
pavlopro, line 2: Perform split operation and save result into variable. Check that lenght is exactly 3. Check that first item is 'cp'

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if file == new_file:
+        return
+
+    with open(file, "r") as file_in, open(new_file, "w") as file_out:
+        for line in file_in:
```
pavlopro, line 7: There is no need iteration over file. Just pass file_in to write of file_out

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1071, opened 2024-07-21
Solution

r0manvynnytskyi:


app/main.py
```diff
@@ -1 +1,22 @@
...
+import shutil
+
+
+def copy_file(command: str) -> None:
+    splited = command.split()
+    cp = splited[0]
+    if cp != "cp":
+        print("Command must be started from 'cp'")
+        return
+    file_name_to_copy = splited[1]
```
vasylhnatiuk, line 10: you should check if command has 3 arguments before it

vasylhnatiuk (changes requested):

vsmutok (changes requested):
Great!
But don't use `shutil`, write your custom logic

vsmutok (approved):
Great!

PR #1072, opened 2024-07-21
Implemented main

OleksiukStepan:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    sep_command = command.split()
```

app/main.py
```diff
@@ -1 +1,5 @@
...
+
+    with open(sep_command[1], "r") as f, open(sep_command[-1], "w") as w:
```
oliasymak24, line 4: `w`  and `f` is a bad variable name for a file

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_copy = command[1]
+    new_file = command[2]
+
+    if file_copy == new_file:
+        return
```
oliasymak24, line 6: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_copy = command[1]
+    new_file = command[2]
```
oliasymak24, line 3: command, old_file, new_file = command.split() will work better)

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1073, opened 2024-07-21
I wrote the code as I expected, but could not test it...

MaksimGolev:


NazarSheremet:
image
Interesting commit name)))

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    source_filename = parts[1]
+    destination_filename = parts[2]
```
NazarSheremet, line 4:
```suggestion
    cmd, source_filename , destination_filename = command.split()
```
This way would look better

NazarSheremet (changes requested):
Need to add check if command is equaled `cp`
image

YuriiSmolii (changes requested):
Check if the destination filepath exists, if not - create.

NazarSheremet (approved):
Good job!

PR #1074, opened 2024-07-21
Solution

andreykolomiec:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        raise ValueError
+    source_file, destination_file = parts[1], parts[2]
+    if source_file == destination_file:
+        return
+    with open(source_file, "r") as file_in, \
+            open(destination_file, "w") as file_out:
```
vsmutok, line 9: better use `()` then `\`

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        raise ValueError
```
```suggestion
    if len(parts) != 3 or parts[0] != "cp":
        raise ValueError("some informative message")
```

app/main.py
```diff
@@ -1 +1,10 @@
...
+    source_file, destination_file = parts[1], parts[2]
+    if source_file == destination_file:
+        return
```
```suggestion
    if source_file == destination_file:
        print("some informative message")
        return
```

vsmutok (changes requested):

app/main.py
andreykolomiec: I have Python version 3.8.

andreykolomiec:
with open(source_file, "r") as file_in, \
            open(destination_file, "w") as file_out:  Try to avoid using line breaking operator \ in the code, place expression inside ()
 I have Python version 3.8. 
    with open(source_file, "r") as file_in:
        with open(destination_file, "w") as file_out:

app/main.py
```diff
@@ -1 +1,14 @@
...
+        raise ValueError(
+            "Invalid command format."
+            "Expected format: cp source_file destination_file"
+        )
+    if source_file == destination_file:
+        raise ValueError(
+            "The source file and the destination file have the same names."
+        )
+    with open(source_file, "r") as file_in:
+        with open(destination_file, "w") as file_out:
```
Dimosphen1, line 13: In case python is a limiting factor, update it to the python 3.10 to place such conditions inside of `()` instead of nesting context managers. Please, resolve an issue mentioned earlier

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, source_file, destination_file, *_ = command.split()
```
vsmutok, line 2: first you should check that the length of `command.split()` is not less than three

vsmutok (changes requested):

pavlopro (approved):
GJ!

PR #1076, opened 2024-07-22
Solution

vldyu:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_to_ls = command.split(", ")
```
Dimosphen1, line 2: There should not be unpacking by comma delimiter, use standard space instead. Besides that, perform unpacking to avoid further indexing
```suggestion
    command_name, input_file, output_file, *_ = command.split()
```
Rewrite code to use this approach

Dimosphen1 (changes requested):

PR #1077, opened 2024-07-22
Implemented solution

dimak20:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    garbage, copy_name, new_name = command.split()
+    if copy_name == new_name:
+        return
```

oliasymak24 (changes requested):

oliasymak24 (approved):
Good job!

PR #1078, opened 2024-07-24
'Solution'

PVS1905:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file1 = command.split()[1]
+    file2 = command.split()[2]
```
vsmutok, line 3: first you should check that the length of `command.split()` is not less than three

app/main.py
```diff
@@ -1 +1,7 @@
...
+
+    if not file1 == file2:
```
vsmutok, line 5: Also Make sure you check that the `cp` input command is correct.

vsmutok (changes requested):

app/main.py
PVS1905: використати try:
except ?

vsmutok (changes requested):
.

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    args = command.split()
+    if len(args) < 3:
+        raise FileNotFoundError("No file name specified.")
+
+    file1 = command.split()[1]
+    file2 = command.split()[2]
```
EdAlekseiev, line 7: Don't use split a few times. You can use variable args here nad unpucking
```suggestion
    file1, file2 = args[1:]
```

app/main.py
EdAlekseiev: Still not fixed

EdAlekseiev (changes requested):

EdAlekseiev (approved):

PR #1079, opened 2024-07-25
Solution

yuliiayedynak:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split(" ")
```
EdAlekseiev, line 2: " " is default value for split
```suggestion
    commands = command.split()
```

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if commands[1] != commands[2]:
```
EdAlekseiev, line 3: You should sheck that command contains 3 elements and first element is "cp"

EdAlekseiev (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split()
```
Dimosphen1, line 2: You could use unpacking and utilizing variables instead of indexing as an alternative approach for better readability
```suggestion
    command_name, input_file, output_file, *_ = command.split()
```

Dimosphen1 (approved):
LGTM, well done! One comment is left for consideration

PR #1084, opened 2024-07-31
Solution

NazarSerdiuk1:


app/main.py
```diff
@@ -1 +1,11 @@
...
+    parts = command.split()
+
+    if len(parts) == 3 and parts[0] == "cp":
+        source_code, destination_code = parts[1], parts[2]
+
+        if source_code != destination_code:
+            with open(source_code) as f, open(destination_code, "wt") as m:
+                m.write(f.read())
+
+            return f"File {source_code} copied to {destination_code}"
```
Dimosphen1, line 11: There is no need to add this string a return, you could omit it

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> str:
+    parts = command.split()
+
+    if len(parts) == 3 and parts[0] == "cp":
+        source_code, destination_code = parts[1], parts[2]
+
+        if source_code != destination_code:
+            with open(source_code) as f, open(destination_code, "wt") as m:
```
Dimosphen1, line 8: Do not use single-letter variable names in the code

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> str:
+    parts = command.split()
+
+    if len(parts) == 3 and parts[0] == "cp":
+        source_code, destination_code = parts[1], parts[2]
+
+        if source_code != destination_code:
```
Dimosphen1, line 7: Try to avoid additional nesting of if conditions in the implementation

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> str:
+    parts = command.split()
```
Dimosphen1, line 2: You could use unpacking and avoid indexing to improve the readability
```suggestion
    command_name, source_file, destination_file, *_ = command.split()
```

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> str:
+    command_name, source_file, destination_file, *_ = command.split()
```
pavlopro, line 2: Split into variable and then check that length is exactly 3
after this you can make unpacking

pavlopro (changes requested):

pavlopro (approved):
Good Job!

PR #1086, opened 2024-08-01
Solution

bknyrik:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    source_file, new_file = command.split(" ")[1:]
```

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
If the command is wrong print the message for a user

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_args = command.split(" ")
```
```suggestion
    command_args = command.split()
```

EdAlekseiev (changes requested):

vsmutok (approved):
Great!

PR #1087, opened 2024-08-02
Solution

dartomOOv:


app/apples.csv
```diff
@@ -0,0 +1,11 @@
...
+buy,10
+buy,13
+supply,17
+buy,10
+supply,5
+supply,80
+buy,39
+supply,56
+buy,32
+buy,11
```
oliasymak24, line 11: You don't need to add these test files

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.split(" ")
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    if "cp" not in files and files.index("cp") != 0:
+        print("Incorrect command to copy file")
+        return
+
+    if files[1] == files[2]:
+        return None
```
oliasymak24, line 9: if a function does not have an explicit return statement, it returns None by default

oliasymak24, line 9: try not to use blank return here, make a backward condition

oliasymak24 (changes requested):

MNDonut (approved):

PR #1088, opened 2024-08-02
solution

HalynaRohatska:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split(" ")
```

app/main.py
```diff
@@ -1 +1,7 @@
...
+    with (
```
oliasymak24, line 3: You seem to have forgotten to check for this condition:
It must do nothing if the user tries to copy the file to file with the same.

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_split = command.split()
+    command_to_read, command_to_write = command_split
```
oliasymak24, line 3: You don't need to separate it between two lines, wi be better to use `command, old_file, new_file = command.split()`

oliasymak24: Don't forget that you also must have "cp" in your command

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    is_cp, command_to_read, command_to_write = command.split()
```
Oleksl888, line 2: what if command split does not return exactly three elements?

app/main.py
```diff
@@ -1 +1,11 @@
...
+    if is_cp != "cp":
+        raise ValueError("The command must start with the command 'cp'")
```
Oleksl888, line 4: No need to raise here - simply combine it with other condition in one if statement and execute code if both are satisfied. If not satisfied - do nothing

app/main.py
```diff
@@ -1 +1,11 @@
...
+
+    if command_to_read != command_to_write:
+        with (
```
Oleksl888, line 7: Would be nice to wrap this in try-except to catch file-related errors

Oleksl888 (changes requested):

app/main.py
HalynaRohatska: but we have condition "Function take only one argument command, that is string with command cp, file name to copy and new file name, separated by spaces."

oliasymak24 (approved):

PR #1090, opened 2024-08-04
Resolve

ihorutkin:


app/file.txt
```diff
@@ -0,0 +1 @@
+Some text
```
oliasymak24, line 1: Yoy don't need to add these test files

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    current_file = command.split(" ")[1]
+    new_file = command.split(" ")[2]
```
oliasymak24, line 3: no need to use space inside split method, it uses space by default

oliasymak24, line 3: you can unpack these variables in the easiest way, also use the more descriptive names

app/main.py
```diff
@@ -1 +1,11 @@
...
+    if current_file != new_file:
+        with (open(current_file, "r") as file_in,
+              open(new_file, "w") as file_out):
+            file_1 = file_in.read()
+            file_out.write(file_1)
+
+
+copy_file("cp file.txt new_file.txt")
```
oliasymak24, line 11: You don't need to add this call of function

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    current_file = command.split(" ")[1]
+    new_file = command.split(" ")[2]
+    if current_file != new_file:
+        with (open(current_file, "r") as file_in,
+              open(new_file, "w") as file_out):
+            file_1 = file_in.read()
+            file_out.write(file_1)
```
oliasymak24, line 8: You can do it in one line

oliasymak24 (changes requested):

app/new_file.txt
```diff
@@ -0,0 +1 @@
+Some text
```
oliasymak24, line 1: Yoy forgot to delete this

oliasymak24 (changes requested):

app/new_file.txt
ihorutkin: Yeah, done)

oliasymak24 (approved):

PR #1091, opened 2024-08-04
Solution

Narberal90:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if not command.startswith("cp "):
+        return
+    read_from, write_to = command[3:].split(" ")
+    if read_from == write_to:
+        return
```
oliasymak24, line 6: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if not command.startswith("cp "):
+        return
+    read_from, write_to = command[3:].split(" ")
```
oliasymak24, line 4: no need to use space inside split method, it uses space by default

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.startswith("cp "):
+        read_from, write_to = command[3:].split()
```
oliasymak24, line 3: Will be better to unpack it at the beginning of the function
command_cp, old_file, new_file = command.split()
to avoid usind slices and calling function startswith
then can just put `if command_cp == "cp" ` that's it

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_cp, read_from, write_to = command.split()
```
Oleksl888, line 2: What if command split returns more then 3 or less then 3 elements?
Also source_file / destination file would be more appropriate names.

app/main.py
```diff
@@ -1 +1,6 @@
...
+
+    if command_cp == "cp" and read_from != write_to:
+        with open(read_from, "r") as file_in, open(write_to, "w") as file_out:
```
Oleksl888, line 5: It is a good practice to wrap working with file in try-except to catch file-related errors

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        raise ValueError(
```
Oleksl888, line 4: In this exercise we recommend to keep code simple. Instead of raising exception here you can structure conditionals in a way that if conditions are met you continue with the rest of code, otherwise nothing will be executed and the program will exit. This will save you couple of lines of code and you may write less instructions

app/main.py
```diff
@@ -1 +1,18 @@
...
+            "Invalid command format. "
+            'Expected format: "cp <source_file> <destination_file>"')
+    source_file = parts[1]
```
Oleksl888, line 7: Use unpacking to get ` source_file and destination_file` - it would be more pythonic approach

app/main.py
```diff
@@ -1 +1,18 @@
...
+    destination_file = parts[2]
+
+    if source_file != destination_file:
+        try:
+            with (open(source_file, "r") as file_in,
+                  open(destination_file, "w") as file_out):
+                file_out.write(file_in.read())
+        except FileNotFoundError:
```
Oleksl888, line 15: FilenotFoundError is a sublcass for OSError, IOError is an alias for OSError - so you can simply use OSError here and it will cover all errors that might happen during file handling

Oleksl888 (changes requested):

app/main.py
oliasymak24: Did you change this comment?

app/main.py
oliasymak24: And this?

oliasymak24 (changes requested):

Oleksl888 (approved):
Well done! But check the last comment about unpacking - this is more elegant way then you did :wink:

PR #1092, opened 2024-08-04
first try

yashkunn:


app/main.py
```diff
@@ -1 +1,25 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
```
oliasymak24, line 5: you can unpack these variables in the easiest way, also use the more descriptive names

app/main.py
```diff
@@ -1 +1,25 @@
...
+
+    if len(parts) != 3 or parts[0] != "cp":
+        raise ValueError("Invalid command format")
+
+    source_file = parts[1]
+    destination_file = parts[2]
+
+    if source_file == destination_file:
+        return
```
oliasymak24, line 14: try not to use blank return here, make a backward condition

PR #1093, opened 2024-08-04
implement py-copy-file

ValeriiaKyr:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
```

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
...
+def copy_file(command: str) -> None:
+    if command.split()[1] != command.split()[2]:
+        with (open(
+                f"{command.split()[1]}",
+                "r"
+        ) as first_file,
+            open(
+                f"{command.split()[2]}",
+                "w"
+        ) as copy_file):
```
MNDonut, line 10: read this message please
https://mate.academy/chats/244801?selectedMessageId=3129519

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command.split()[1] != command.split()[2]:
+        with (open(
+                f"{command.split()[1]}",
+                "r"
+        ) as first_file,
+            open(
+                f"{command.split()[2]}",
+                "w"
```
MNDonut, line 9: you split the same line 4 times. If it's a big string, it'll take some time, let's save splitted command in a variable and reuse it ;)

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_command = command.split()
```

app/main.py
```diff
@@ -1 +1,17 @@
...
+    name_firs_file = list_command[1]
+    name_copy_file = list_command[2]
+
+    if name_firs_file != name_copy_file:
+        with (
+            open(
+                f"{command.split()[1]}",
+                "r"
```
oliasymak24, line 10: No need to put this "r" in a new line the same fot other case

app/main.py
oliasymak24: command, old_file, new_file = command.split()

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, name_firs_file, name_copy_file = command.split()
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if name_firs_file != name_copy_file:
```
Oleksl888, line 3: you should also have to check if command starts with `cp`

app/main.py
```diff
@@ -1 +1,12 @@
...
+        with (
```
Oleksl888, line 4: would be nice to wrap it in try-except to catch file-related errors

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
...
+            open(
+                f"{name_firs_file}", "r"
+            ) as first_file,
```
Oleksl888, line 7: source_file / destination file would be more appropriate names imo

app/main.py
```diff
@@ -1 +1,30 @@
...
+
+
+def copy_file(command: str) -> None:
+    if len(command.split()) == 3:
+        try:
+            command, source_file, destination_file = command.split()
+            if command != "cp":
+                raise NotRightCommand("Incorrect command")
+            if source_file == destination_file:
+                raise SameNamesError("File names are the same!")
```
MNDonut, line 16: I recommend to replace this code by one "if" that will check all 3 conditions simultaneously and otherwise nothing happens

MNDonut (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands, source_file, destination_file = command.split()
```
MNDonut, line 2: What if I am a bad user and will enter this command?
`cp file1.txt file2.txt file3.txt file4.txt`
Your unpacking will fail because of too many elements

MNDonut (changes requested):

MNDonut (approved):

PR #1094, opened 2024-08-05
Task Solution

achalaidiuk:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+class FileNotFound(Exception):
```
oliasymak24, line 1: in Python we have bult-in exception FileNotFoundError

app/main.py
```diff
@@ -1 +1,17 @@
...
+    pass
+
+
+def copy_file(command: str) -> None:
+    operation, file_name, destination = command.split()
+
+    if operation != "cp" or file_name == destination:
+        print("File error")
```
oliasymak24, line 9: > It must do nothing in case the user is trying to copy file to file with the same name 


but try not to use blank return here, make a backward condition

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    operation, file_name, destination = command.split()
+
+    try:
+        if operation != "cp" or file_name == destination:
+            print("File error")
```
oliasymak24, line 6: You don't need this print, better first check file_name != destination and operation == "cp" and call your context maneger without any `else`

oliasymak24 (changes requested):

LiudmylaKulinchenko (approved):
Great job 🤠

PR #1095, opened 2024-08-05
Solution

mihavryliuk:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    com = command.split()
```

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    action, source_file, destination_file = command.split()
+
+    if action != "cp" or len(command.split()) != 3:
```
Oleksl888, line 4: would make sense to check len of command split before you actually split and unpack - otherwise you might get error when unpacking

app/main.py
```diff
@@ -1 +1,16 @@
...
+        raise ValueError("Command in wrong format")
```
Oleksl888, line 5: no need to raise errors here simply rewrite that if command is wrong or file names are the same it does nothing and if everything correct it will proceed with executing rest of logic

app/main.py
```diff
@@ -1 +1,16 @@
...
+    if source_file != destination_file:
+        try:
+            with (
+                open(source_file, "r") as file_in,
+                open(destination_file, "w") as file_out
+            ):
+                file_out.write(file_in.read())
+        except FileNotFoundError:
```
Oleksl888, line 13: OSError would be more appropriate exception here - it will cover all file-related errors

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3 or parts[0] != "cp":
+        return
+
+    _, source_file, destination_file = parts
+
+    if source_file == destination_file:
```
Oleksl888, line 9: Still not optimal - you have two  unnecessary return instructions. Try to follow this:
check if len of command is correct, unpack command, check if command is correct, proceed with the rest of logic.
Think how you can write your conditional statements so that it if they are true you continue with execution and if they are not your code does nothing. this way you will not need to add extra return instructions

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3:
```
Oleksl888, line 4: This is the pattern we are looking for 
```
if len(parts) == 3:
   rest of logic here

**Review by Oleksl888 — CHANGES_REQUESTED** [review]

File: app/main.py [hunk 7]
```diff
@@ -1 +1,13 @@
...
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) == 3:
+        action, source_file, destination_file = parts
+
+        if action == "cp" and source_file != destination_file:
+            try:
+                with open(source_file, "r") as file_in:
+                    with open(destination_file, "w") as file_out:
```
LiudmylaKulinchenko, line 10: You can actually write it in one line! Let's do this last fix 💪 Looks good overall 😌

LiudmylaKulinchenko (changes requested):

app/main.py
mihavryliuk: line to long

oliasymak24 (approved):

PR #1096, opened 2024-08-05
Politov Solution

Alexandr-Politov:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    command_list = command.split()
```
oliasymak24, line 3: you can unpack these variables in the easiest way, also use the more descriptive names

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    if command_list[1] == command_list[2] or command_list[0] != "cd":
+        print("do nothing")
+        return None
+
+    with (open(command_list[1], "r") as file_out,
+          open(command_list[2], "w") as file_to):
+        print("copy file")
```
oliasymak24, line 11: Delete this unnecessary print

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    command_list = command.split()
+
+    if command_list[1] == command_list[2] or command_list[0] != "cd":
+        print("do nothing")
+        return None
```
oliasymak24, line 7: By defult if you just call `return` you will get None, no need to put `return None` but anyway try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    command_list = command.split()
+
+    if command_list[1] == command_list[2] or command_list[0] != "cd":
+        print("do nothing")
```
oliasymak24, line 6: No need for this print too

oliasymak24 (changes requested):

oliasymak24 (approved):
Good job!

PR #1097, opened 2024-08-05
Solution of py-copy-file

AndriiHamasa:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    words = command.split()
```

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+    if words[1] == words[2]:
+        return
```
oliasymak24, line 5: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,10 @@
...
+
+    with open(words[1]) as file_in, open(words[2], "w") as file_out:
+        content = file_in.read()
+
+        file_out.write(content)
```
oliasymak24, line 10: You can do it in one line

oliasymak24 (changes requested):

app/test.txt
```diff
@@ -0,0 +1,2 @@
+hello
+world
```
oliasymak24, line 2: Delete this test file

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    _, first_name, second_name = command.split()
+
+    if first_name != second_name:
```
oliasymak24, line 4: I think will be also good to add to this `if` if you have "cp" in your command

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, first_name, second_name = command.split()
```

app/main.py
```diff
@@ -1 +1,6 @@
...
+
+    if first_name != second_name and command == "cp":
+        with open(first_name) as file_in, open(second_name, "w") as file_out:
```
Oleksl888, line 5: Would make sense to wrap this in try-except to catch file-related errors

Oleksl888 (changes requested):

PR #1098, opened 2024-08-05
first

panichqa:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_cp = command.split()
+    if len(command_cp) != 3 or command_cp[0] != "cp":
+        return
+
+    cp, source, destination = command_cp
```
oliasymak24, line 6: Better to move it to line 2

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+    if source == destination:
+        return
```
oliasymak24, line 9: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_cp = command.split()
+    if len(command_cp) != 3 or command_cp[0] != "cp":
+        return
```
You can do it with this check together `if source == destination`

oliasymak24 (changes requested):

app/main.py
LiudmylaKulinchenko: You still have blank returns. Fix throughout the code

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    if len(command_parts) != 3 or command_parts[0] != "cp":
```
LiudmylaKulinchenko, line 3: Let's check only length here, and check  if first elements equals cp after we unpack varialbes to make it more self-explanatory

LiudmylaKulinchenko (changes requested):

app/main.py
Oleksl888: Please fix as per previous comments - remove unnecessary prints and returns and nest logic under if condition

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1099, opened 2024-08-05
First attempt

dpiuro:


app/main.py
```diff
@@ -1 +1,38 @@
...
+class CustomFileNotFoundError(CopyFileError):
+    pass
+
+
+class FileCopyError(CopyFileError):
+    pass
+
+
+def copy_file(copy_command: str) -> None:
+    command, source_file, destination_file = copy_command.split()
```
Oleksl888, line 22: Would make sense to check that len of command is equal to 3 first. Otherwise your program will crash during unpacking

app/main.py
```diff
@@ -1 +1,38 @@
...
+
+    if command != "cp" or len(copy_command.split()) != 3:
+        raise InvalidCommandError("Invalid command format")
+    if source_file == destination_file:
+        raise SameFileError(f"{source_file} "
```
Oleksl888, line 27: As per task requirement:

> It must do nothing in case the user is trying to copy file to file with the same name.

So it would make sense to use a single if statement where you check if len is correct, if command is correct and if file name is not the same. That is much easier then making multiple ifs and raising exceptions in them

app/main.py
```diff
@@ -1 +1,38 @@
...
+                            f"and {destination_file} files are the same.")
+    try:
+        with (
+            open(source_file, "r") as file_in,
+            open(destination_file, "w") as file_out
+        ):
+            file_out.write(file_in.read())
+    except FileNotFoundError:
```
Oleksl888, line 35: OSError would cover most errors related to work with file, so you don't need to create your own exception classes here

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        parts = command.split()
+
+        if (
+                len(parts) != 3
```
Oleksl888, line 6: I suggest to refactor this:
1. Try block should be covering only operations with file - OSError should be used in as an exception we are trying to catch. So no need to reraise OSError - you may catch it and print a message that error occured during file handling.
2. If statement should be rewritten backwards - so you don't have to create extra return instruction. So if all conditions are met - execute the rest of code, if not then script will exit 
3. Use unpacking to get `source_file` and `destination_file` - it would be more pythonic approach

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1101, opened 2024-08-05
Solution

Bogdan-Rylov:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
```

app/main.py
```diff
@@ -1 +1,14 @@
...
+    if (
+            len(command_list) != 3
+            or command_list[0] != "cp"
+            or command_list[1] == command_list[2]
+    ):
+        return
```
oliasymak24, line 8: try not to use blank return here, make a backward condition

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    file1_name = command_list[1]
+    file2_name = command_list[2]
+    with open(file1_name) as file1, open(file2_name, "w") as file2:
```
oliasymak24, line 12: file1, file2 is not very good practice for variables name

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, source_file_name, destination_file_name = command.split()
```
Oleksl888, line 2: if command is wrong your unpacking may crash

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if command == "cp" and source_file_name != destination_file_name:
+        with (
```
Oleksl888, line 4: would make sense to wrap below block in try except to catch file-related errors

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        command, source_file_name, destination_file_name = command.split()
+    except ValueError:
+        print("Wrong command format")
+    else:
+        if command == "cp" and source_file_name != destination_file_name:
+            with (
```
Oleksl888, line 8: does your code handle file-related errors?

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        command, source_file_name, destination_file_name = command.split()
+    except ValueError:
```
Oleksl888, line 4: nice solution here

Oleksl888 (changes requested):

oliasymak24 (approved):

PR #1103, opened 2024-08-06
'Solution'

ABaidin:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None | ValueError:
+    if command.split()[0] != "cp" or len(command.split()) != 3:
+        raise ValueError("Invalid command format")
+
+    original, copy = command.split()[1], command.split()[2]
```
oliasymak24, line 5: you can unpack these variables in the easiest way, also use the more descriptive names

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if original == copy:
+        return
```
oliasymak24, line 7: try not to use blank return here, make a backward condition

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1104, opened 2024-08-06
Solution

Ivan-Shakhman:
My version of solution

### DIFF
--- a/app/main.py
+++ b/app/main.py
@@ -1,1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    temp_lst = command.split()
+    if len(temp_lst) != 3 or any(value for value in temp_lst) is None:
+        raise ValueError("incorrect data. You must specify 'command' \
+        'file name' and 'name of the copy to create'")
+    if temp_lst[0] != "cp":
+        raise ValueError(f"Unknown command {temp_lst[0]}")
+    if temp_lst[1] == temp_lst[2]:
+        raise ValueError("You can not create new file with same name")
+    with open(temp_lst[1], "r") as file_in, open(temp_lst[2], "w") as file_out:
+        file_out.write(file_in.read())
+    print(f"File {temp_lst[1]} was successfully copied to {temp_lst[2]}")


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    temp_lst = command.split()
+    if len(temp_lst) != 3 or any(value for value in temp_lst) is None:
```
Oleksl888, line 3: As per task requirement:

>     It must do nothing in case the user is trying to copy file to file with the same name.

So it would make sense to use a single if statement where you check if len is correct, if command is correct and if file name is not the same. That is much easier then making multiple ifs and raising exceptions in them. Also maybe you can use unpacking instead of accessing elements by index

app/main.py
```diff
@@ -1 +1,12 @@
...
+        raise ValueError("incorrect data. You must specify 'command' \
+        'file name' and 'name of the copy to create'")
+    if temp_lst[0] != "cp":
+        raise ValueError(f"Unknown command {temp_lst[0]}")
+    if temp_lst[1] == temp_lst[2]:
+        raise ValueError("You can not create new file with same name")
+    with open(temp_lst[1], "r") as file_in, open(temp_lst[2], "w") as file_out:
```
Oleksl888, line 10: Would be nice to wrap this in try-except to catch file-related errors

app/main.py
```diff
@@ -1 +1,12 @@
...
+        file_out.write(file_in.read())
+    print(f"File {temp_lst[1]} was successfully copied to {temp_lst[2]}")
```
Oleksl888, line 12: This print is unnecessary

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    comm, file_name, copy_name = command.split()
+    if len(command.split()) != 3 or comm != "cp" or file_name == copy_name:
+        return
```

oliasymak24 (changes requested):

oliasymak24 (changes requested):
)

oliasymak24 (approved):

PR #1105, opened 2024-08-06
implement py-core-file

Maksym-Turenko:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3 or parts[0] != "cp":
```
Oleksl888, line 4: All below code be simplified to one if statement. Remove unnecessary return and raises. Instead check if command is correct i one if and then nest the code under this condition if it is satisfied, if it's not satisfied - don't do anything.

app/main.py
```diff
@@ -1 +1,13 @@
...
+        raise ValueError("Invalid command format")
+
+    src_file, dest_file = parts[1], parts[2]
```
Oleksl888, line 7: it would be more pythonic way to use unpacking on iterable

app/main.py
```diff
@@ -1 +1,13 @@
...
+
+    if src_file == dest_file:
+        return
+
+    with open(src_file, "r") as file_in, open(dest_file, "w") as file_out:
```
Oleksl888, line 12: Would be a good idea to wrap this in try-except to catch file-related errors

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) == 3 and parts[0] == "cp":
+        src_file, dest_file = parts[1], parts[2]
```
Oleksl888, line 5: would be more pythonic to use unpacking here

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+        if src_file != dest_file:
+            try:
+                with (
+                    open(src_file, "r") as file_in,
+                    open(dest_file, "w") as file_out
+                ):
+                    file_out.write(file_in.read())
+            except (FileNotFoundError, IOError) as e:
```
Oleksl888, line 14: FilenotFoundError is a sublcass for OSError, IOError is an alias for OSError - so you can simply use OSError here and it will cover all errors that might happen during file handling

Oleksl888 (changes requested):

Oleksl888 (approved):
Well done!

PR #1106, opened 2024-08-06
Solution

pashawarganov:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+class IncorrectCommand(Exception):
+    pass
+
+
+def copy_file(command: str) -> None:
+    command = command.split(" ")
```
Oleksl888, line 6: No need to split by space , it is used by default

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if len(command) != 3 or "cp" not in command:
+        raise IncorrectCommand("Incorrectly written command")
```
Oleksl888, line 8: This can be simplified to one if statement. Remove unnecessary return and raises. Instead check if command is correct i one if and then nest the code under this condition if it is satisfied, if it's not satisfied - don't do anything.

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if command[1] == command[2]:
```
Oleksl888, line 9: you can unpack command instead of accessing elements by indexes

app/main.py
```diff
@@ -1 +1,13 @@
...
+        return
+
+    with open(command[1], "r") as file_in, open(command[2], "w") as file_out:
```
Oleksl888, line 12: Would be nice to wrap this in try-except to catch file-related errors

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, filename1, filename2 = command.split()
+
+    if filename1 != filename2 and command == "cp":
+        try:
+            with open(filename1, "r") as file_in, \
```
oliasymak24, line 6: we don't use "\" to move to a new line

oliasymak24 (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if filename1 != filename2 and command == "cp":
+        try:
+            with open(filename1, "r") as file_in:
+                with open(filename2, "w") as file_out:
+                    file_out.write(file_in.read())
+        except FileNotFoundError as e:
+            print(e)
+
+
+copy_file("cp abc bca.txt")
```
oliasymak24, line 13: Delete this test code

oliasymak24 (approved):

PR #1107, opened 2024-08-06
solution

Sergi0bbb:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split(" ")
```
LiudmylaKulinchenko, line 2: you can unpack these variables in the easier way, also use the more descriptive names

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, old_file, new_file = command.split()
```
Oleksl888, line 2: what if command split does not return 3 elements as you expect? 
Also from checklist:

> Make sure you check that the cp input command is correct.

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if old_file != new_file:
+        with (
```
Oleksl888, line 4: Would be nice to wrap this in try-except to catch file-related errors

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    if len(command_parts) != 3:
+        raise ValueError(f"Expected 3 arguments, got {len(command_parts)}")
+
+    command, old_file, new_file = command_parts
```
Oleksl888, line 6: Missing check that command must start with `cp`

app/main.py
```diff
@@ -1 +1,15 @@
...
+    if old_file == new_file:
```
Oleksl888, line 7: Your approach with raising Exceptions looks well structured however in this task  I would suggest for this task to use conditional statements to check that command is correct. You can structure conditionals in a way that if conditions are met you continue with the rest of code, otherwise nothing will be executed and the program will exit. This will save you couple of lines of code and you may write less instructions

app/main.py
```diff
@@ -1 +1,15 @@
...
+        raise ValueError(f"File {new_file} already exists")
+
+    try:
+        with open(old_file, "r") as file_in:
+            with open(new_file, "w") as file_out:
+                file_out.write(file_in.read())
+    except FileNotFoundError:
```
Oleksl888, line 14: OSError would cover most errors related to work with file, not only when file not found. Instead of reraising you can print that something went wrong during file handling

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split()
+    if len(command_parts) != 3:
```
Oleksl888, line 3: Can be more simple with less code:
```
if len(command_parts) != 3:
    unpack command parts to variables
    -> check other conditions in single if statement
        -> put rest of code under this conditional

Oleksl888 (changes requested):

PR #1108, opened 2024-08-06
Task

Poznanskyi:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, first_name, second_name = command.split()
```

app/main.py
```diff
@@ -1 +1,6 @@
...
+    if first_name != second_name and command == "cp":
+        with (open(first_name, "r") as file_in,
```
Oleksl888, line 4: would be nice to wrap it in try-except to catch file-related errors

Oleksl888 (changes requested):

app/main.py
Oleksl888: you need to check if split-command has exactly three elements otherwise your program will crash during unpacking

Oleksl888: I'd suggest to apply `try` block only to file-related operations - starting from when you open file.

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command, first_name, second_name = command.split()
+    try:
+        if first_name != second_name and command == "cp":
+            with (open(first_name, "r") as file_in,
+                    open(second_name, "w") as file_out):
+                file_out.write(file_in.read())
+    except FileNotFoundError:
```
Oleksl888, line 8: OSError would cover more file and system related errors

Oleksl888 (changes requested):

LiudmylaKulinchenko (changes requested):
`you need to check if split-command has exactly three elements otherwise your program will crash during unpacking` -- you didn't fix this comment

LiudmylaKulinchenko (approved):
Good job 🌷

PR #1109, opened 2024-08-06
Solution

mnezbrytska:


app/main.py
```diff
@@ -1 +1,26 @@
...
+    new_str = command.split()
+    if len(new_str) < 3:
+        return "argument is not correct"
+    file_first = new_str[1]
+    file_second = new_str[2]
+    cmd = new_str[0]
+    if cmd != "cp":
+        return "Command is not correct"
+
+    if file_first == file_second:
```
Oleksl888, line 12: All above code can be simplified to one or two if statements. So remove unnecessary return and pass instructions. Try to check if len is correct, if command is correct and if file name is not the same and then put rest of the code under this condition if it is satisfied, if it's not satisfied - don't do anything.  Also maybe you can use unpacking instead of accessing elements by index

app/main.py
```diff
@@ -1 +1,26 @@
...
+                open(file_first, "r") as first,
+                open(file_second, "w") as second
+            ):
+                for line in first:
+                    second.write(f"{line}")
+        except OSError as e:
+            print(f"Error {e}")
+
+
+copy_file("cp file.txt new_file.txt")
```
Oleksl888, line 26: this shouldn't be here

Oleksl888 (changes requested):

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+
+def copy_file(command: str) -> None:
+    new_str = command.split()
+
+    file_first = new_str[1]
+    file_second = new_str[2]
+    cmd = new_str[0]
```
oliasymak24, line 7: Not use indexes here, you can unpack it in easiest way)

app/main.py
```diff
@@ -1 +1,21 @@
...
+
+    if file_first != file_second and cmd == "cp" and len(new_str) == 3:
+
+        try:
+            with (
+                open(file_first, "r") as first,
+                open(file_second, "w") as second
+            ):
+                for line in first:
```
oliasymak24, line 16: try to do it without loop

oliasymak24 (changes requested):
Your tests were failed(

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+
+def copy_file(command: str) -> None:
+    new_str = command.split()
+    cmd, file_first, file_second = new_str
```
MNDonut, line 4: what if the command is "cp file1.txt file2.txt file3.txt"? unpacking will fail

MNDonut (changes requested):

app/main.py
mnezbrytska: yes, you are right, but I can't unpacking after if. So, I need return to my first way with index or divide if to two parts?

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+
+def copy_file(command: str) -> None:
+    new_str = command.split(" ", 2)
```
Oleksl888, line 3: you don't need to set separator and max split here - it will split by spaces by default . Then you don't need to set maxsplit because it may return incorrect data - it will always try to split in 3 parts
```
command = "This command is wrong"
command.split(" ", 2)
['This', 'command', 'is wrong'] # 3 elements
command.split()
['This', 'command', 'is', 'wrong'] # 4 elements

File: app/main.py [hunk 7]
```diff
@@ -1 +1,15 @@
...
+    cmd, file_first, file_second = new_str
+
+    if len(new_str) == 3 and file_first != file_second and cmd == "cp":
```
Oleksl888, line 6: you need to check `len(new_str)` before you split otherwise program may crash during unpacking

app/main.py
```diff
@@ -1 +1,15 @@
...
+
+        try:
+            with (
+                open(file_first, "r") as first,
```
Oleksl888, line 10: use more descriptive names instead of first/second

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+
+def copy_file(command: str) -> None:
+    new_str = command.split(" ", 2)
+    cmd, file_first, file_second = new_str
```
Oleksl888, line 4: Use more appropriate names such as `source_file/destination_file`

Oleksl888 (changes requested):

app/main.py
Oleksl888, line 3: `new_str ` is not a very good name either

LiudmylaKulinchenko (approved):
Good job 🔥

PR #1112, opened 2024-08-08
implement py-copy-file

dkn3r:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    data = command.split()
+    file1, file2 = data[1], data[2]
```
vasylhnatiuk, line 3: you should check if data has 3 arguments before it

vasylhnatiuk (changes requested):

PR #1113, opened 2024-08-09
Solution

PavelTym:


app/main.py
```diff
@@ -1 +1,19 @@
...
+def copy_file(command: str) -> None:
+    copy_info = command.split()
+
+    source_file = copy_info[1]
+    target_file = copy_info[2]
+
+    if not copy_info:
+        print("Command 'cp' is empty")
+
+    if len(copy_info) != 3:
```
pavlopro, line 10: This should be checked first of all
Please correct message. It doesn't relates to this check

app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> None:
+    copy_info = command.split()
+
+    source_file = copy_info[1]
+    target_file = copy_info[2]
+
+    if not copy_info:
```
pavlopro, line 7: Why did you have this check if above you access values of copy_info?

pavlopro (changes requested):

app/main.py
pavlopro: This was not resolved.
Imagine that copy_info has only one item. so in this case you will encountering an error.

pavlopro (changes requested):

app/main.py
pavlopro: Good, than you could ensure that the [0] element is exactly 'cp' =)

pavlopro (changes requested):

PR #1114, opened 2024-08-11
Solution

VovkOL:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command.split()
+    if file_names[1] == file_names[2]:
```
vasylhnatiuk, line 3: you should to check if file_names has 3 arguments before it

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1,7 +1,15 @@
 def copy_file(command: str) -> None:
     file_names = command.split()
+
+    if len(file_names) != 3:
```
vasylhnatiuk, line 4: also check if command is correct (cp)

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1120, opened 2024-08-14
solution

dvitalijb:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split(" ")
+    file_in_name = command_parts[1]
+    file_out_name = command_parts[2]
```
NazarSheremet, line 4: This way looks better 
```suggestion
    command, file_in_name, file_out_name = command.split()
```

NazarSheremet (changes requested):
Fix suggestion and add check if command equals `cp`
image

YuriiSmolii (approved):
LGTM

PR #1121, opened 2024-08-15
Finished

Oleksa-32:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    file1, file2 = parts[1], parts[2]
```
vasylhnatiuk, line 3: you should check if command has 3 arguments

vasylhnatiuk (changes requested):

PR #1122, opened 2024-08-16
My solution

lavryniuk-dev:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if not (command.startswith("cp ")):
+        return
```
LiudmylaKulinchenko, line 3: You should also check if there are 3 words in the command

app/main.py
```diff
@@ -1 +1,12 @@
...
+
+    cmd, file1, file2 = command.split()
+    if (file1 == file2
+            or not (file1.endswith(".txt"))
+            or not (file2.endswith(".txt"))):
```
LiudmylaKulinchenko, line 8: fix formatting

LiudmylaKulinchenko (changes requested):

LiudmylaKulinchenko (approved):
Looks good 🌸

PR #1125, opened 2024-08-20
Solution

fposlushnyi:


LLkaia (approved):
GJ

But in future, if some operations like this:
```
    path_in = os.path.join(os.path.dirname(__file__), file_in)
    path_out = os.path.join(os.path.dirname(__file__), file_out)
```
can be skipped because arguments are not correct, it is better to firstly do validation and then do such kind of operations since they can be resource consuming.

PR #1129, opened 2024-08-25
Solution

ip-ilya:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
```
pavlopro, line 1: First of all you should split and check that length is exactly 3, then you could upack it into variables

pavlopro (changes requested):

app/main.py
```diff
@@ -1 +1,16 @@
...
+    if len(command.split()) == 3:
+        command_name = command.split()[0]
+        source_file_name = command.split()[1]
+        destination_file_name = command.split()[2]
```
Dimosphen1, line 5: Avoid performing `split` operation for 3 times here, use unpacking instead
```suggestion
        command_name, source_file_name, destination_file_name = command.split()
```

Dimosphen1 (changes requested):

pavlopro (approved):
GJ!

PR #1136, opened 2024-08-27
copy-file

AndreyHMkr:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    part = command.split()
```
LLkaia, line 2:
```suggestion
    parts = command.split()
```

LLkaia: or `argument_list`

app/main.py
```diff
@@ -1 +1,15 @@
...
+    if len(part) != 3 or part[0] != "cp":
+        return
+    source = part[1]
+    destination = part[2]
+    if source == destination:
+        return
+    try:
+        with open(source, "r") as old_name, open(destination, "w") as new_file:
```
LLkaia, line 10: since it's not a name but opened file object
```suggestion
        with open(source, "r") as old_file, open(destination, "w") as new_file:
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+            new_file.write(old_name.read())
+    except FileNotFoundError:
+        print(f"File {source} does not exist.")
+    except Exception as e:
+        print(str(e))
```
LLkaia, line 15: `e` calling will return string so no reason to convert it

LLkaia (changes requested):

LLkaia (approved):

PR #1138, opened 2024-08-28
Solution

arsen-arutiunov:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")[1:]
```
vasylhnatiuk, line 2: you have to check if command has 3 arguments before it

vasylhnatiuk (changes requested):

PR #1144, opened 2024-09-02
Solution

idubina:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    files = command.strip("cp ").split()
+    if files[0] != files[1]:
```
LLkaia, line 3: you need to check if `cp` is not missed and there are two arguments after that

app/main.py
```diff
@@ -1 +1,5 @@
...
+        with open(files[0], "r") as file_in, open(files[1], "w") as file_out:
```
LLkaia, line 4: you are doing indexing twice so mb it worth to save file names in different variables instead of getting them by index each time

LLkaia (changes requested):

LLkaia (changes requested):
I think, `cd` command stands for another purpose...

LLkaia (approved):
Perfect!

PR #1146, opened 2024-09-03
Solution

bdnnk:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file, file_copy = command.split(" ")[1:3]
```
LLkaia, line 2: What if cp missed? what if len of args < 3?

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    if file == file_copy:
+        return
```
LLkaia, line 5: try to avoid redundant return

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    with open(file, "r") as f, open(file_copy, "w") as fc:
+        content = f.read()
+        fc.write(content)
```
LLkaia, line 9: write it in one liine

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+
+command = "cd andrii Suprun.txt"
+
+copy_file(command)
```
LLkaia, line 14: remove these lines

LLkaia (changes requested):
Additionally, remove `andrii` file from PR

app/main.py
```diff
@@ -1 +1,13 @@
...
+
+    if len(command_parts) < 3 or command_parts[0] != "cp":
+        raise ValueError(
+            "Invalid command format. ""Expected 'cp file copy_file'."
+        )
+
+    file, copy_file = command_parts[1], command_parts[2]
+
+    if file != copy_file:
+        with open(file, "r") as f, open(copy_file, "w") as fc:
```
LLkaia, line 12:
```suggestion
        with open(file, "r") as original_f, open(copy_file, "w") as destination_f:
```

LLkaia (approved):
Looks good!

Consider name variable more specific in future.

PR #1147, opened 2024-09-03
"Solution."

ShiftRomero:


app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_names = command.split(" ")
+    if (file_names[1] != file_names[2]
```
vasylhnatiuk, line 3: you have to check if command has 3 arguments if not you should write message for user

vasylhnatiuk (changes requested):

PR #1148, opened 2024-09-04
solution

uzlss:


app/main.py
```diff
@@ -1 +1,6 @@
-# write your code here
+def copy_file(command: str) -> None:
+    file_name = command.split()[1]
```
vasylhnatiuk, line 2: use split() only one time

vasylhnatiuk: and check if command has 3 argument before this

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1156, opened 2024-09-10
Solution

MahdaArtem-py:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        return
```
vasylhnatiuk, line 4: print message for user that command is not correct

app/main.py
```diff
@@ -1 +1,10 @@
...
+    name_to_copy = parts[1]
```
```suggestion
    name_to_copy, new_name = parts[1], parts[2]
```

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1162, opened 2024-09-18
Solution

anastasiiakalini:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    file, new_file = command.split()[1], command.split()[2]
+    if file != new_file:
+        shutil.copy(file, new_file)
```
Dimosphen1, line 7: Use context manager and corresponding read and write operations to perform copy

Dimosphen1 (changes requested):

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp_command, file, new_file = command.split()
```
pavlopro, line 2: Check that lenght of result of split is exactly 3

app/main.py
```diff
@@ -1 +1,9 @@
...
+    if cp_command != "cp":
```
pavlopro, line 3: Check that `file` is not equal to `new_file`

app/main.py
```diff
@@ -1 +1,9 @@
...
+        raise ValueError(f"The first argument must be 'cp'! "
+                         f"Instead '{cp_command}' provided.")
+
+    with open(file, "r") as file_in, open(new_file, "w") as file_out:
+        content = file_in.read()
+        file_out.write(content)
```
pavlopro, line 9: No need to have this `content` variable. Directly pass the result of read method into write method

pavlopro (changes requested):

pavlopro (approved):
GJ

PR #1170, opened 2024-09-27
copy file

Lev2098:


app/main.py
```diff
@@ -1 +1,16 @@
...
+
+    if len(parts) != 3 or parts[0] != "cp":
+        print("some informative message")
+        return
+
+    source_file, destination_file = parts[1], parts[2]
+
+    if source_file == destination_file:
+        print("some informative message")
+        return
```
YukihiroSM, line 12: Mentor ment, that you should print some information about what's going on.  Use some descriptive prints in this case,please

YukihiroSM (changes requested):
Else looks pretty good to me!

app/main.py
Lev2098: Ohh, okey

PR #1172, opened 2024-09-28
copy file

dmytrik:


app/main.py
```diff
@@ -1 +1,14 @@
...
+    # we are cheking on correct data input
+    if command != "cp" or file_name == new_file_name:
+        return
+    # we are opening current file and creat new file
+    # using mode "w" and write down there data from
+    # current_file
+    with (open(file_name, "r") as file_in,
+          open(new_file_name, "w") as file_out):
+        for line in file_in.readlines():
+            file_out.write(line)
```
AnyoneClown, line 14: You don't need to use `for loop` here. We can simply `read()` all content from file first and then write it in 1 moment.

Check `read()` method for this and fix it

app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command_line: str) -> None:
+    (command,
+     file_name,
+     new_file_name) = command_line.split(" ")
```
AnyoneClown, line 4: `.split()` is splitting spaces by default, so we don't need to pass any `args` to it in this case

AnyoneClown (changes requested):

AnyoneClown (approved):
Good job!

PR #1173, opened 2024-09-29
Solution

PonomaryovNikita:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split(" ")
```
pavlopro, line 2: `command.split()`

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if split_command[1] != split_command[2]:
```
pavlopro, line 3: Check that length of split result is exactly 3
Check that first element is 'cp'

pavlopro (changes requested):

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
+    first_check = len(split_command) == 3
```
pavlopro, line 3: no need to have those variables

pavlopro (changes requested):

LLkaia (approved):
LGTM

PR #1174, opened 2024-09-30
Solution

KKvladys:


dmytrik (commented):
1.split is splitting spaces by default, you don't need to pass any args there.
2.you can don't use shutil (new_file.write(file.read())
3. you can make 1 common condition(if cm != "cp" or file == new_file: return)

dmytrik (approved):

AnyoneClown (approved):
LGTM!

PR #1176, opened 2024-10-01
'Solution'

Obusker:


app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_files(file_names: str) -> None:
+    command, file, new_file = file_names.split()
```
vasylhnatiuk, line 2: you have to check if input has 3 arguments before it

vasylhnatiuk (changes requested):

app/main.py
vasylhnatiuk, line 2: you have to check if input has 3 arguments **before** this line

vasylhnatiuk (changes requested):

vasylhnatiuk (approved):

PR #1177, opened 2024-10-02
Created function to copy file and checked it

Mxbely:


app/main.py
```diff
@@ -1 +1,23 @@
...
+from contextlib import contextmanager
+
+
+@contextmanager
+def open_file(name: str, method: str) -> None:
+    func = open(name, method)
+    try:
+        yield func
+    finally:
+        func.close()
```
AnyoneClown, line 10: You can do without a separate `open_file` function by using the built-in context manager `open` for this

app/main.py
```diff
@@ -1 +1,23 @@
...
+
+
+def copy_file(command: str) -> None:
+    command = command.split()
+    if command[0] != "cp":
+        return
+    if command[1] == command[2]:
+        return
```
AnyoneClown, line 18: You can implement it in 1 `if` condition

AnyoneClown (changes requested):

AnyoneClown (approved):
Good, but check comment

PR #1179, opened 2024-10-03
solution

IvanGuculyak:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, name, new_name = command.split()
+    if len(command.split()) != 3:
+        print("Wrong command!")
```
```suggestion
    _command = command.split()
    if len(_command) != 3:
        print("Wrong command!")
        return 
```

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, name, new_name = command.split()
```
vsmutok, line 2: before unpacking check that the length of command.split() is 3

vsmutok (changes requested):

vsmutok (changes requested):

vasylhnatiuk (approved):

PR #1181, opened 2024-10-06
Solution

tsukimi1394:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_name, source_filename, target_filename, *_ = command.split(" ")
```
vsmutok, line 2: First, check that the length of command.split() is equal to three.

```suggestion
    command_name, source_filename, target_filename, *_ = command.split()
```

vsmutok (changes requested):

PR #1182, opened 2024-10-07
Solution

apmaz:


pavlopro (changes requested):
Please remove `file.txt` and `new_file.txt` from PR

README.md
```diff
@@ -14,7 +14,7 @@ Example:
 ```python
 copy_file("cp file.txt file.txt")  # Does nothing
 
-copy_file("cp file.txt new_file.txt")
+copy_file("cp 11.txt new_file.txt")
```
pavlopro, line 17: do not modify this file

app/main.py
```diff
@@ -1 +1,5 @@
-# write your code here
+def copy_file(command: str) -> None:
+    cp, file, new_file = command.split()
```
pavlopro, line 2: First you should split to check that length is exactly 3

```
split_result = command.split()
if len(command.split()) == 3:
    cp, file, new_file = split_result
```


then make other checks

app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_result = command.split()
+    if len(command.split()) == 3:
```
LLkaia, line 3: u can use split_result instead of calling command.split() again

LLkaia (changes requested):

LLkaia (approved):
LGTM

PR #1183, opened 2024-10-08
Solution

vladyslav-tmf:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        print("Invalid command format.")
+        return
+    source_file, target_file = parts[1], parts[2]
+    if source_file == target_file:
+        print("Source and target files are the same. Nothing to copy.")
```
VladyslavBon, line 8: It must do nothing in case the user is trying to copy file to file with the same name

VladyslavBon (approved):

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        print("Invalid command format.")
+        return
+    source_file, target_file = parts[1], parts[2]
+    if source_file == target_file:
```
Arsen-hrynevych, line 7: The check if source_file == target_file: simply returns if the files are identical. This avoids unnecessary work, but it might be more helpful to print a message indicating that the source and target files are the same, which could prevent confusion during debugging.

app/main.py
```diff
@@ -1 +1,17 @@
...
+        return
+
+    try:
+        with (open(source_file, "r") as file_in,
+              open(target_file, "w") as file_out):
+            file_out.write(file_in.read())
+    except FileNotFoundError:
+        print(f"File {source_file} not found.")
+    except Exception as error:
+        print(f"An error occurred: {error}")
```
Arsen-hrynevych, line 17: add more handlers for exceptions

Arsen-hrynevych (approved):
ok

PR #1185, opened 2024-10-08
Solution

SchOleh:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if command[:2] == "cp":
```
vasylhnatiuk, line 2: you have to check if command has 3 arguments before it

vasylhnatiuk (changes requested):

vasylhnatiuk (changes requested):
use command.split() only once

vasylhnatiuk (approved):

PR #1188, opened 2024-10-09
copy-file solution

94nj111:


app/main.py
```diff
@@ -1 +1,10 @@
...
+def copy_file(command: str) -> None:
+    command = command.split(" ")
+    cp = command[0]
+    file_from = command[1]
+    file_to = command[2]
+    if cp == "cp" and file_from != file_to:
+        with open(file_from, "r") as f:
+            content = f.read()
+        with open(file_to, "w") as f:
+            f.write(content)
```
Mxbely, line 10: It is bad practice, but i think it is not mistake. 
You can use two contexts managers together, it will be more easy to read and smaller code
image

Mxbely (approved):
Not bad, but check my comments, please

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split(" ")
```
AnyoneClown, line 2: `.split()` method works the same by default. If u are not passing anything iside - it will split by space

```suggestion
    command = command.split()
```

app/main.py
AnyoneClown, line 10: Fix names. prefer `new_file`, `old_file`

AnyoneClown (changes requested):

app/main.py
Nattalli: that comment left

Nattalli (changes requested):

Arsen-hrynevych (approved):
ok

PR #1189, opened 2024-10-09
Solution

Icemann007:


app/main.py
```diff
@@ -1 +1,7 @@
-# write your code here
+def copy_file(command: str) -> None:
+    ls = command.split(" ")
+    if ls[0] == "cp":
```
vasylhnatiuk, line 3: you have to check if ls has 3 arguments before this line

vasylhnatiuk (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    ls = command.split()
+    if len(ls) == 3:
+        if ls[0] == "cp":
+            if ls[1] == ls[2]:
+                print("Does nothing")
+            with open(ls[1], "r") as file_in, open(ls[2], "w") as file_out:
+                file_out.write(file_in.read())
```
```suggestion
    ls = command.split()
    if len(ls) == 3 and ls[0] == "cp" and ls[1] != ls[2] :
        with open(ls[1], "r") as file_in, open(ls[2], "w") as file_out:
            file_out.write(file_in.read())
    else:
        print("command should be in this format <cp file.txt new_file.txt>")
```

vsmutok (changes requested):

PR #1198, opened 2024-10-14
solution

mshmygel:


SerhiiK05 (approved):
If I remember correctly, if you use WITH there in no need to close file. It's automatically closed after you done with it. 
Code works - GJ!

AnyoneClown (approved):
Good job!

PR #1206, opened 2024-10-15
Solution

molodsh1y1:


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    copy_command, source_file, destination_file = command.split(" ")
+
+    if copy_command == "cp":
+        if source_file == destination_file:
+            return
+
+        with (open(source_file, "r") as file_in,
+              open(destination_file, "w") as file_out):
```
Den-k0, line 9: You can use two contexts managers simultaneously by separating them by a comma:
```
    ...
```

app/main.py
```diff
@@ -1 +1,11 @@
...
+            for line in file_in:
+                file_out.write(line)
```
Den-k0, line 11: Can be written more simply and without iterating in "for" loop for all lines. Like a:
`file_out.write(file_in.read())`

Den-k0 (changes requested):

Den-k0 (approved):
Great work!

Den-k0 (approved):

PR #1208, opened 2024-10-15
Develop

Bondzik-S:


app/main.py
dxrrkwm, line 1: you can make 3 variables out of command.split to make it more readable

dxrrkwm, line 1: also you can remove for loop and use source.read

dxrrkwm (changes requested):

dxrrkwm (approved):
looks good i guess!

AnyoneClown (approved):
Good job!

PR #1225, opened 2024-10-23
Solution

serg121192:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
+        raise ValueError("Wrong copy command input")
+    cmd = command.split()[0]
+    data_file = command.split()[1]
+    data_file_copy = command.split()[2]
```
LLkaia, line 6: Please, achieve needed result doing split only once.

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if len(command.split()) != 3:
+        raise ValueError("Wrong copy command input")
+    cmd, data_file, data_file_copy = command.split()
```
LLkaia, line 4: Do split only once. Result assign to list, then check length, then unpack it.

LLkaia (changes requested):

PR #1230, opened 2024-10-27
solution

JRuiy:


app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    file_names = command.replace("cp ", "").split(" ")
```
vsmutok, line 5: please first check that the command was valid

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,8 @@
...
+    if file_names[0] == file_names[1] or not command.startswith("cp"):
```
vsmutok, line 6: Please  first make sure you check that the input string contains 3 elements separated by space, and has `cp` on a first place.

vsmutok (changes requested):

app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+import shutil
+
+
+def copy_file(command: str) -> None:
+    commands = command.split(" ")
```
```suggestion
    commands = command.split()
```

`" "` - by default

app/main.py
```diff
@@ -1 +1,12 @@
...
+    if (
+            len(commands) != 3
+            or not commands[0] == "cp"
```
```suggestion
            or commands[0] != "cp"
```

app/main.py
```diff
@@ -1 +1,12 @@
...
+            or commands[1] == commands[2]
+    ):
+        return
```
vsmutok, line 11:
```suggestion
        print("some informative message, so that the user understands that something went wrong")
        return
```

vsmutok (changes requested):

vsmutok (approved):
Good Job!

PR #1237, opened 2024-11-01
py-copy-file

konstantinou77:


Alkaponees (changes requested):
Recommend adding exception and conditions. For example, check length of command and check if user use right command. Exception on existing file

PR #1239, opened 2024-11-03
Solution

whatmode:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_attributes = [command_param
+                          for command_param in command.split()
+                          if command_param not in ["cp", ""]]
```
NazarSheremet, line 4: I don't like this check; it looks illogical because the requirement is to verify whether the first parameter is equal to `cp` and whether it contains two path parameters. However, this check doesn't verify either of those

NazarSheremet (changes requested):

app/main.py
whatmode: I see, do you want me do add also a check to verify that we have 2 parameters and specific command?
This code is used just to get command attributed in a list

whatmode:
I pushed new solution, please check

Alkaponees (changes requested):
Recommend adding exceptions such as FileNotFound and IOError

Alkaponees (approved):
LGMT!

PR #1247, opened 2024-11-08
Solution

88victory88:


Alkaponees (changes requested):
Please add checking for length of command and start of parts

88victory88:
> Please add checking for length of command and start of parts

Please, explain what you mean "start of parts" I can understand it. And the length of command can't be less than six symbols, isn't it?

app/main.py
```diff
@@ -1 +1,8 @@
-# write your code here
+
+def copy_file(command: str) -> None:
+    splited_command = command.split(" ")
```
YuriiSmolii, line 3: check if the parsed command started from "cp"

YuriiSmolii (changes requested):

Alkaponees (approved):
Good Job!

PR #1262, opened 2024-11-21
Solution

fredisson11:


Alkaponees (changes requested):
Add condition to check length of command and start command with cp

Alkaponees (approved):
LGMT!

PR #1264, opened 2024-11-21
copy function impemented

Beliar24:


Alkaponees (changes requested):
Add condition to check length of command and if command satrt with cp

Alkaponees (approved):
LGMT!

PR #1265, opened 2024-11-22
Solution

trinidaa:


Alkaponees (changes requested):
Add conditions to check length of command and its start

Alkaponees (approved):
LGMT!

PR #1274, opened 2024-11-29
Solution_v1.0

Rozdorozhnii:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split(" ")
+
+    if (len(split_command) == 3
```
Nattalli, line 4: raise an error if some of the conditions are not true

Nattalli: also, you can create variables for split_command[1] and split_command[2] as you reuse it

Nattalli (changes requested):

Nattalli (approved):

PR #1287, opened 2024-12-12
Created function

shafran-andriy:


app/main.py
```diff
@@ -1 +1,12 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
```
pavlopro, line 2: 1. by default it split by space, so you can simplify this
2. Before accessing values it would be good to check that len of command_list is exactly 3
3. check that first command is cp

pavlopro (changes requested):

app/main.py
shafran-andriy: fixed

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if len(command_list) == 3 and command_list[0] == "cp":
```
pavlopro, line 3: `... and command_list[-1] != command_list[-2]:`

app/main.py
```diff
@@ -1 +1,13 @@
...
+        if command_list[-1] == command_list[-2]:
+            pass
+        else:
+            with open(command_list[-2],
+                      "r") as file_in, open(command_list[-1],
+                                            "w") as file_out:
+                file_out.write(file_in.read())
+
+
+copy_file("cp file.txt new_file.txt")
```
pavlopro, line 13: remove this code

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_list = command.split()
+    if len(command_list) == 3 and command_list[0] == "cp":
+        if command_list[-1] == command_list[-2]:
+            pass
+        else:
+            with open(command_list[-2],
+                      "r") as file_in, open(command_list[-1],
```
pavlopro, line 8: format it be make it user readable

pavlopro (changes requested):

app/main.py
shafran-andriy: fixed

app/main.py
shafran-andriy: removed

app/main.py
shafran-andriy: done

pavlopro (approved):

PR #1299, opened 2024-12-15
Solution

IvankaKuzin:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+class CommandError(Exception):
+    pass
+
+
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
+
+    if command_list[0] == "cp":
```
4ikkamony, line 8: Task states that in addition ot == "cp" you gotta check if command string you received has only 3 elements, and also if file_in and file_out are the same function should do nothing.

app/main.py
```diff
@@ -1 +1,15 @@
...
+        with (
+            open(command_list[1], "r") as file_in,
+            open(command_list[2], "w") as file_out,
```
4ikkamony, line 11: currently, if file names are the same, line 11 will immediately overwrite the file, and your reader variable later will be an empty string. Which will be then written to the file

app/main.py
```diff
@@ -1 +1,15 @@
...
+        ):
+            reader = file_in.read()
+            file_out.write(reader)
+    raise CommandError
```
4ikkamony, line 15: Also, error is always raised, no matter what happens

4ikkamony (changes requested):
Cool idea with CommandError
But there are a few things to fix, good luck!

PR #1308, opened 2024-12-16
Solution

bogdAAAn1:


app/main.py
```diff
@@ -1 +1,18 @@
...
+
+    if file_names[1] == file_names[2]:
+        return
+
+    try:
+        with (
+            open(file_names[1], "r") as file_in,
+            open(file_names[2], "w") as file_out
+        ):
+            file_out.write(file_in.read())
```
Professor-Douel, line 14: Now it`s not a problem, but when you need to copy very big file it can be a problem, because when you start copy, it takes much memory, better copy the file by rows. 
For example:
`for line in file_in:
    file_out.write(line)`

Professor-Douel (approved):

PR #1321, opened 2024-12-17
solution copy file

SummerKatrin:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split = command.split(" ")
+    command = split[0]
+
+    if command != "cp":
```
yelizaveta-zh, line 5: I'm not sure we need to do this check

yelizaveta-zh (approved):
Everything looks perfectly written, except for one point - maybe you're doing too much checking.

app/main.py
```diff
@@ -2,0 +1,15 @@
...
+        return
+
+    file1 = split[1]
```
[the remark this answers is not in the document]
L1nk3rrr: agreed

app/main.py
```diff
@@ -2,0 +1,15 @@
...
+    file2 = split[2]
```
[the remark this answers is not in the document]
L1nk3rrr: agreed

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split = command.split(" ")
```
L1nk3rrr, line 2: You can assign all your variables here!
`command, source..., dest... = command.split(" ")`

L1nk3rrr (changes requested):
Look at my comments and some from AI

app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_parts = command.split(" ")
```
petrykivd, line 2: You can simplify the unpacking in the way Valentyn wrote to you earlier.

app/main.py
```diff
@@ -1 +1,21 @@
...
+    command = command_parts[0]
+
+    if command != "cp":
+        return
+
+    source_file = command_parts[1]
+    destination_file = command_parts[2]
+
+    if source_file == destination_file:
+        return
```
petrykivd, line 12: You can use 1 if for return:
if source_file == destination_file or command != "cp":
    return

petrykivd (changes requested):

petrykivd (approved):
Nice! 😎

PR #1323, opened 2024-12-18
solute

MaksVakulenko:


bodiakof (approved):
Neat solution!

app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(cp: str) -> None:
+    parts = cp.split(" ")
+    old = parts[1]
+    new = parts[2]
```
L1nk3rrr, line 4: you can assign variables into one line

`command, old, new = input.split(" ")`

Also old and new is a bad names, maybe try smth. like source and destination>?

app/main.py
```diff
@@ -1 +1,10 @@
...
+    if old == new:
+        return
+    else:
```
L1nk3rrr, line 7: no needed else, because in `if` you returning, so interrupt function call

L1nk3rrr (changes requested):
Well done! Please check AI and my comments for improvement of your code

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(cp: str) -> None:
+    parts = cp.split(" ")
```
petrykivd, line 2: You can simplify the unpacking in the way Valentyn wrote to you earlier.

app/main.py
```diff
@@ -1 +1,11 @@
...
+    if len(parts) != 3 or parts[0] != "cp":
+        raise Exception
+    source = parts[1]
+    destination = parts[2]
+    if source == destination:
+        return
```
petrykivd, line 8: Simplify and combine two conditions into one.
if source == destination or command != "cp":
    return

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(cp: str) -> None:
```
petrykivd, line 1: Maybe command is better naming for this variable? cp is just a part of command

petrykivd (changes requested):

petrykivd (approved):
Well done! 💪

PR #1324, opened 2024-12-18
Solution

bodiakof:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts: list = command.split(maxsplit=2)
+
+    if len(parts) != 3 or parts[0] != "cp":
+        return
+
+    source_file: str = parts[1]
+    target_file: str = parts[2]
+
```
petrykivd, line 9: You can simplify the unpacking in next way:
command, source, target= command.split()

app/main.py
```diff
@@ -1 +1,20 @@
...
+    if source_file == target_file:
+        return
```
petrykivd, line 11: Simplify and combine two conditions into one.
if source == target or command != "cp":
return

petrykivd (changes requested):

L1nk3rrr (approved):
Well done!

PR #1349, opened 2025-01-02
task done

Mykola-Osolinskyi:


app/main.py
```diff
@@ -1 +1,10 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
Nattalli, line 2: for better reading, you can do something like this, but that is optional:
```
    parts = command.split()
    cp = parts[0]
    old_file = parts[1]
    new_file = parts[2]

File: app/main.py [hunk 2]
```diff
@@ -1 +1,10 @@
...
+    if len(parts) != 3 or parts[0] != "cp":
+        raise ValueError("Invalid command format")
+
+    with open(parts[1], "r") as file_in, open(parts[2], "w") as file_out:
+        if file_in == file_out:
+            return
+    input_data = file_in.read()
+    file_out.write(input_data)
```
Nattalli, line 10: do that in the `with` block

Nattalli (changes requested):

app/main.py
Mykola-Osolinskyi: done

app/main.py
Mykola-Osolinskyi: done

Nattalli (approved):

PR #1394, opened 2025-02-03
solution

Baranotik15:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    data = command.split()
```
oliasymak24, line 2: you can unpack these variables in the easier way, also use the more descriptive names
Example: command, old_file, new_file = data.split()

app/main.py
```diff
@@ -1 +1,14 @@
...
+
+    if data[1] == data[2]:
+        return
+
+    try:
+        with open(data[1], "r") as f_in, open(data[2], "w") as f_out:
```
oliasymak24, line 8: `f_out` is a bad variable name for a file

app/main.py
oliasymak24, line 1: Try to use `argparse` module here — it would be easier, more readable, and you won't have to reinvent the wheel

oliasymak24 (changes requested):

oliasymak24 (approved):

PR #1409, opened 2025-02-10
implement copy_file function

yarzhuravlov:


app/main.py
oliasymak24, line 1: Try to use `argparse` module here — it would be easier, more readable, and you won't have to reinvent the wheel

oliasymak24 (changes requested):
Your solution also works but recomendation to try)

oliasymak24 (approved):

PR #1414, opened 2025-02-12
'Solution'

JohnyJDX:


app/new_file
```diff
@@ -0,0 +1 @@
+asfasfasfasf
```
Nattalli, line 1: you can remove that from the PR

app/main.py
```diff
@@ -1 +1,9 @@
-# write your code here
+def copy_file(command: str) -> None:
+    new_file = command.split()
+    test1 = new_file[1]
+    test2 = new_file[2]
```
Nattalli, line 4: you can do something like this:
```python
    name_in, name_out = command.split()[1:]
```

Nattalli (changes requested):

Nattalli (approved):

PR #1434, opened 2025-02-26
Completed task copy-file

Yurkiv313:


app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+def copy_file(command: str) -> None:
+    part_of_command = command.split()
```
Y-Havryliv: not the best variable name, `part_of_command` suggests that it's a singular, but there are many parts of the command. So it's better to change it

app/main.py
```diff
@@ -1 +1,18 @@
...
+    if len(part_of_command) != 3 or part_of_command[0] != "cp":
+        print("Incorrect file format")
```
Y-Havryliv: For some simple scripts, it may be OK, but it's better to use `raise` when something goes wrong.
Also, the message is not correct and misleads the reader.

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+def copy_file(command: str) -> None:
+    part_of_command = command.split()
+    if len(part_of_command) != 3 or part_of_command[0] != "cp":
```
Y-Havryliv: also `part_of_command[0]` feels "magical". Create a variable as well and name it somehow

app/main.py
```diff
@@ -1 +1,18 @@
...
+        print("Incorrect file format")
+        return
+
+    file_1, file_2 = part_of_command[1], part_of_command[2]
+
+    if file_1 == file_2:
+        print("Incorrect file name")
```
Y-Havryliv: wrong message here. The name of the file is correct, but the files are actually the same. So it's better to change it as well

Y-Havryliv (changes requested):

Boyok1337 (approved):
Good job, but remember about `raise` instead of print

PR #1443, opened 2025-03-02
'Solution'

AndyStarGitHub:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_words = command.split()
+    if len(command_words) < 3:
+        return
```
Yurkiv313, line 4: you can include a print with a message before return to explain the problem.
so it will be better to understand.

Yurkiv313 (approved):

PR #1444, opened 2025-03-02
Implement copy_file function

PrimeGlorious:


app/main.py
```diff
@@ -1 +1,14 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
+
+    try:
+        if split_command[0] != "cp" or split_command[1] == split_command[2]:
+            return None
```
Yurkiv313, line 11: Before return, you can add a print to explain the reason for exiting the function.
Also, instead of accessing elements by index, it’s better to use unpacking:
command_name, source_file, destination_file = split_command

Yurkiv313 (approved):

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    split_command = command.split()
```
oliasymak24, line 2: you can unpack these variables in the easier way,

app/main.py
oliasymak24: Good point

app/main.py
```diff
@@ -1 +1,16 @@
...
+        if (len(split_command) != 3
+                or split_command[0] != "cp"
+                or split_command[1] == split_command[2]):
+            return None
+
+        with (open(split_command[1], "r") as current_file,
+              open(split_command[2], "w") as new_file):
+            current_file_content = current_file.read()
+
+            new_file.write(current_file_content)
```
oliasymak24: You can do it in one line

oliasymak24 (changes requested):

Y-Havryliv (changes requested):
Why didn't you fix previous comments?

Y-Havryliv (approved):

PR #1523, opened 2025-04-02
Solution

farzamgt:


app/main.py
```diff
@@ -1 +1,24 @@
...
+    source_file = parts[1]
+    target_file = parts[2]
+
+    if source_file == target_file:
+        return
+
+    if not os.path.exists(source_file):
+        return
+
+    with (
```
Tolik-code, line 19: You can use shutil.copyfile to do this in one method, it would be like this "shutil.copyfile(source_file_name, target_file_name)"

app/main.py
```diff
@@ -1 +1,24 @@
...
+import os
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+
+    if len(parts) != 3 or parts[0] != "cp":
+        return
+
+    source_file = parts[1]
```
Tolik-code, line 10: There can use comprehension to get params from array, would be like this "_, source_file, target_file = parts"

Tolik-code (approved):
Good job!

PR #1624, opened 2025-07-02
'Solution'

DanilHamster:


app/main.py
```diff
@@ -1 +1,19 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
AnyoneClown: Use unpacking here

Example:
```
var1, var2, var3 = command.split()
```

AnyoneClown (changes requested):

AnyoneClown (approved):
LGTM

PR #1640, opened 2025-07-07
Solution

Artcrafterrra:


app/main.py
```diff
@@ -1 +1,17 @@
...
+import os
+
+
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3 or parts[0] != "cp":
+        return
+    if not os.path.exists(parts[1]):
+        return
+    if parts[1] == parts[2]:
```
Orixara, line 10: I think there are a lot of if operations  and you can separate parts to: cmd, file_name, new_file_name = parts, it will be better to understand and instead: 
if not os.path.exists(parts[1]):
        return
you can use try/except FileNotFoundError in block with(open())

Orixara (approved):

PR #1643, opened 2025-07-08
solution py copy file

PavelSemenikhin:


Vskesha (approved):
Good concise solution. 
The code is simple and understandable. It also meets all the requirements.

PR #1646, opened 2025-07-09
Solution

Kostenikov:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    try:
+        action, source, destination = command.split()
+    except ValueError:
+        return
+
+    if action != "cp":
```
Sashoxxx: I think u'd add check for trying to copy file to file with the same
name

Sashoxxx (approved):

Kostenikov:
I dont need to validate command strint because if i try to unpack when we have less then 2 element or more then 3, we will raise ValueError and call return.

PR #1648, opened 2025-07-09
copy file

Salmon-eat:


app/main.py
```diff
@@ -1 +1,11 @@
...
+    division = command.split(" ")
+    if len(division) == 3 and division[0] == "cp":
+        if division[1] != division[2]:
+            try:
+                with (open(division[1], "r") as file_in,
+                      open(division[2], "w") as file_out):
+                    read_text = file_in.read()
+                    file_out.write(read_text)
+            except Exception:
+                pass
```
dimashevchukk: use different types of exceptions and print errors specified

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    division = command.split(" ")
+    if len(division) == 3 and division[0] == "cp":
+        if division[1] != division[2]:
+            try:
+                with (open(division[1], "r") as file_in,
+                      open(division[2], "w") as file_out):
+                    read_text = file_in.read()
+                    file_out.write(read_text)
```
dimashevchukk: would be problem with big files, can use loop instead

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    division = command.split(" ")
+    if len(division) == 3 and division[0] == "cp":
+        if division[1] != division[2]:
+            try:
+                with (open(division[1], "r") as file_in,
+                      open(division[2], "w") as file_out):
```
dimashevchukk: don't like indexes much, unpacking to different values would be more readable

app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+def copy_file(command: str) -> None:
+    division = command.split(" ")
```
dimashevchukk: space is default param for split

app/main.py
```diff
@@ -1 +1,11 @@
...
+    if len(division) == 3 and division[0] == "cp":
```
dimashevchukk: print error specified

dimashevchukk (changes requested):

app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        cp, old_file, new_file = command.split()
+        for_len = cp, old_file, new_file
+        if len(for_len) == 3 and cp == "cp":
```
dimashevchukk, line 5: would be better to check these thing separately, with printing problem specified

app/main.py
```diff
@@ -1 +1,15 @@
...
+            if old_file != new_file:
+                try:
+                    with (open(old_file, "r") as file_in,
+                          open(new_file, "w") as file_out):
+                        for line in file_in:
+                            file_out.write(line)
+                except FileNotFoundError:
```
dimashevchukk, line 12: not the only exception that can appear, also catch IsADirectoryError, PermissionError, general Exception in the end

app/main.py
```diff
@@ -1 +1,15 @@
...
+                    print("FileNotFoundError:")
+    except ValueError:
+        print("ValueError")
```
dimashevchukk, line 15: bad practice, see AI comment

dimashevchukk (approved):

PR #1651, opened 2025-07-09
develop

MorsImmortalis05:


SensibleN00B (changes requested):
"copy_file("cp file.txt file.txt")  # Does nothing"
if two identical file names are passed to the function, then nothing needs to be done

PR #1660, opened 2025-07-09
Solution

DenisPrudius:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.strip().split()
+
+    if len(parts) != 3 or parts[0] != "cp":
+        print("Invalid command format")
+        return
```
dimashevchukk: better separate this two checks and specify what is wrong

app/main.py
```diff
@@ -1 +1,17 @@
...
+
+    cp, source, target = parts
+
+    if source == target:
+        return
+
+    try:
+        with open(source, "r") as file_in, open(target, "w") as file_out:
+            file_out.write(file_in.read())
```
dimashevchukk: would be a problem with copying big files, use loop instead

app/main.py
```diff
@@ -1 +1,17 @@
...
+    except FileNotFoundError:
+        print(f"File '{source}' not found.")
```
dimashevchukk, line 24: catch other exceptions too, such as isdirectory, permissionerror

dimashevchukk (changes requested):

app/main.py
DenisPrudius: Done

app/main.py
DenisPrudius: Done

dimashevchukk (approved):

PR #1661, opened 2025-07-09
created copy function

trellle:


app/main.py
```diff
@@ -1 +1,13 @@
...
+    if len(params) != 3 or params[0] != "cp" or params[1] == params[2]:
+        return
+    try:
+        file1 = open(params[1], "r")
+    except FileNotFoundError as e:
+        print(e)
+        return
+    content = file1.read()
+    file1.close()
+    file2 = open(params[2], "w")
```
Plovuezz, line 12: why you are not using context manager? And also you handle errors only if you open first file! 
You should rewrite it using two contexts managers simultaneously by separating them by a comma:

then you don`t need to close it manually

Plovuezz (approved):
I am approving but you have to fix this

app/main.py
trellle: ok, I'll remember it

PR #1801, opened 2025-10-29
'Solution'

Vorongor:


app/main.py
```diff
@@ -1 +1,17 @@
...
+
+    if input_name == output_name:
+        return
+
+    try:
+        with (open(input_name, "r") as file_in,
+              open(output_name, "w") as file_out):
+            for line in file_in:
+                file_out.write(line)
+    except Exception as e:
```
DonDmytron: specify the exception you may receive

app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
DonDmytron: use unpacking and reduce the number of checks

DonDmytron (changes requested):

DonDmytron (approved):
good )

PR #1813, opened 2025-11-04
done

VadymVynokur:


app/main.py
```diff
@@ -1 +1,21 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command = command.split()
+    if len(command) != 3:
+        print("invalid command")
+        return
+
+    if command[1] == command[2]:
```
Vorongor: A better option may be using unpacking to create variables and reuse them.

app/main.py
```diff
@@ -1 +1,21 @@
...
+        print("same files")
+        return
+
+    if command[0] == "cp":
```
Vorongor: A slightly cleaner structure may be if combine this check with the previous.

Vorongor (approved):
Your code is correct and readable. Improvements are mostly about: readability (unpack args), slightly cleaner structure.

PR #1818, opened 2025-11-06
Solution

leemonch1k-space:


app/main.py
```diff
@@ -1 +1,13 @@
...
+def copy_file(command: str) -> None:
+    command_list = command.split(" ")
+    if len(command_list) < 3 or command_list[0] != "cp":
+        return
+
+    try:
+        with open(command_list[1], mode="r") as file_read:
+            file_data = file_read.readlines()
+
+        with open(command_list[2], mode="w") as file_write:
```
ostapshpeha: with open(split[1], "r") as file_in, open(split[2], "w") as file_out:
        pass
simple way to open one file and write another

ostapshpeha (approved):

app/main.py
leemonch1k-space: thx

PR #1829, opened 2025-11-09
Soullution

Iventyk:


AlexMishKo (changes requested):
works fine! but if u use recomendation from readme file and combine both opens in one line 
(with open(file_name, "r") as file, open(new_file_name, "w") as new_file )- there will be no need in double try: except block. <--need to redo
And idk if AI will approove this or not, but if annotation is None, and ur function return False... mb it better to just return. With no False. But im not shure^^

PR #1843, opened 2025-11-10
Copy file

ohorodnykostap:
Copy file


app/main.py
```diff
@@ -1 +1,26 @@
-# write your code here
+class FileCopier:
```
CookieTanuki, line 1: Class implementation is not necessary and not required in README.md but overall great job!

CookieTanuki (approved):

PR #1890, opened 2025-12-29
Solution

tarasmosiichuk01-ship-it:


app/main.py
```diff
@@ -1 +1,15 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
```
DaTrEvTeR, line 2: It's better to move each part of the command into a separate variable after validation and work with them separately. This makes the code easier to read and reduces the number of operations required to retrieve a list attribute by index.
```py
command, source_file_path, destination_file_path = parts
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+    if len(parts) != 3:
+        return
+    if parts[0] != "cp":
```
DaTrEvTeR: Combine this 2 if statements in one block
```py
if (command != "cp") or (source == destination):
    return
```

app/main.py
```diff
@@ -1 +1,15 @@
...
+        return
+    file_in = parts[1]
+    file_out = parts[2]
+    if file_in == file_out:
+        return
+    try:
+        with open(file_in, "r") as f_in, open(file_out, "w") as f_out:
```
DaTrEvTeR: Use more descriptive variable names
For example:
```py
with open(source_file_path, "r") as source_file, open(destination_file_path, "w") as copied_file:
    ...
```

DaTrEvTeR (changes requested):

app/main.py
tarasmosiichuk01-ship-it: Added changes

vsmutok (approved):
Great!

PR #1924, opened 2026-01-21
Implement copy_file function

nazarkravets2007:
Implemented copy_file function that works like Linux cp command.

- Parses command string "cp src dst"
- Does nothing if src and dst are the same
- Copies full file content from source to destination


nazarkravets2007 (commented):
Add cp command validation in copy_file

PR #1937, opened 2026-01-31
Solution

PlayWithDen4ik:


app/main.py
```diff
@@ -1 +1,16 @@
...
+    split_command = command.split()
+    if len(split_command) != 3:
+        return
+    cmd, source, dest = split_command
+    if cmd != "cp":
+        return
+    if source == dest:
+        return
+    try:
+        with open(source, "r") as file1, open(dest, "w") as file2:
```
vsmutok: This is a good implementation, but the variable names `file1` and `file2` could be more descriptive. According to checklist item [#2](https://github.com/mate-academy/py-copy-file/pull/2), it's better to use names that indicate the purpose of the variable, such as `source_file` and `destination_file` or `file_in` and `file_out` as suggested in the task description.

app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    if not command:
+        return
```
vsmutok, line 3: Also, I recommend not just returning nothing here, but raising an error with a message that describes what went wrong.
For example: `raise Exception("An invalid command format was provided")`

vsmutok (changes requested):

app/main.py
PlayWithDen4ik: Yes, I understand this approach, and it is appropriate for projects, but not for tasks of this kind, since the tests will not be passed.

vsmutok (approved):
Great!

PR #1967, opened 2026-02-15
copy file

KostiantinPastukh:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(some_file: str) -> None:
+    if len(some_file.split()) < 3:
+        return
+
+    cp, old_file, new_file = some_file.split()
```
vsmutok, line 5: Avoid calling `split()` twice – You currently use `some_file.split()` multiple times. Call it once, store the result in a variable (e.g., `parts = some_file.split()`), and reuse it.

vsmutok (changes requested):

PR #2059, opened 2026-04-01
solved

pawunder1992:


app/main.py
```diff
@@ -1 +1,18 @@
...
+def copy_file(command: str) -> None:
+    try:
+        command = command.split()
+        cur_file = command[1]
+        new_file = command[2]
+    except Exception:
+        return
+    if cur_file == new_file or len(command) != 3 or command[0] != "cp":
+        return
+    else:
```
VadimPusiak788: You don't have to use a block **else** because you already return the result in the **if** statement

app/main.py
```diff
@@ -1 +1,18 @@
...
+        try:
+            with open(cur_file), open(new_file, "w"):
+                pass
+        except FileNotFoundError:
+            return
+        else:
```
VadimPusiak788: You can use a single **with** statement, because trying to open a file that doesn’t exist raises an exception. This also avoids duplicating the with block
`try:
    with open(cur_file) as file_in, open(new_file, "w") as file_out:
except FileNotFoundError:
    return
`

app/main.py
```diff
@@ -1 +1,18 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
```
VadimPusiak788: You don’t need to use a try/except statement here. It would be better to split the string first, then use an if statement to check the length of the resulting command list.

VadimPusiak788 (changes requested):

app/main.py
pawunder1992: done

app/main.py
pawunder1992: done

app/main.py
pawunder1992: done

VadimPusiak788 (approved):
Well done

PR #2079, opened 2026-04-11
Implemented copy function

PavelPereverzev1:


alisahoys (approved):
Correct command parsing, code is readable and structured. Missing same-file check.

app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    commands = command.split(" ")
+    if commands[0] == "cp" and len(commands) == 3:
+        try:
+            if commands[1] == commands[2]:
```
VadimPusiak788: Please move this conditional out of the try-except block

app/main.py
```diff
@@ -1 +1,13 @@
...
+                return
+            with (
+                open(commands[1]) as source,
+                open(commands[2], "w") as target
+            ) :
+                target.write(source.read())
+        except FileNotFoundError:
+            pass
```
VadimPusiak788: Replace pass with return

VadimPusiak788 (changes requested):

app/main.py
```diff
@@ -1 +1,13 @@
...
+    commands = command.split(" ")
+    if commands[0] == "cp" and len(commands) == 3:
+        if commands[1] == commands[2]:
+            return
+        try:
+            with (
+                open(commands[1]) as source,
+                open(commands[2], "w") as target
+            ) :
+                target.write(source.read())
```
AnyoneClown, line 11: In line 11, you open the target with "w", which truncates (erases) the file immediately. If the paths refer to the same file, you will wipe your source data before you can read it.

AnyoneClown (changes requested):

VadimPusiak788 (approved):

PR #2082, opened 2026-04-11
Develop

speg142:


app/main.py
```diff
@@ -1 +1,17 @@
-# write your code here
+def copy_file(command: str) -> None:
+    try:
+        command_name, source, destination = command.split()
+    except ValueError:
+        return
```
vsmutok, line 6: You should print a message here
so that users can understand what went wrong and why they did not get the result.

vsmutok (changes requested):
You should print a message here
so that users can understand what went wrong and why they did not get the result.

PR #2147, opened 2026-05-12
'Solution'

dem1hh:


app/main.py
```diff
@@ -1 +1,20 @@
-# write your code here
+def copy_file(command: str) -> None:
+
+    parts_file = command.split()
+    if len(parts_file) != 3:
+        return
```
vsmutok, line 6: You are returning an empty value. For our task this is ok, but imagine this in a real-world example: a user runs it, gets nothing, and doesn’t understand what went wrong.

So I would recommend that before returning nothing, you return some informative message instead.

vsmutok (changes requested):

vsmutok (approved):
Great!

PR #2255, opened 2026-07-30
add solution

Kuper18:


app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+def copy_file(command: str) -> None:
+    command_data = command.split()
+
+    if len(command_data) < 3:
+        return
+
+    cp, current_file, new_file = command_data
```
DonDmytron, line 7: 'cp' - bad naming, 'command' could be better

app/main.py
```diff
@@ -1 +1,22 @@
...
+
+    if cp != "cp":
```
DonDmytron, line 9: You can use 1 check with 'and'

DonDmytron (approved):

PR #2260, opened 2026-08-06
feature: implemented copy file method

oleksii-tkachuk:


app/main.py
```diff
@@ -1 +1,22 @@
-# write your code here
+from typing import Any
+
+
+def copy_file(command: str) -> Any:
+    try:
+        command_split = command.split(" ")
```
DonDmytron, line 6: you can use unpacking here and make one check instead of three

DonDmytron (approved):

PR #2276, opened 2026-08-13
Solution

sofiia-pohuliailo388:


app/main.py
```diff
@@ -1 +1,23 @@
...
+        return
+
+    source = elements[1]
+    destination = elements[2]
+
+    if source == destination:
+        return
+
+    try:
+        with open(source, "r") as file_in:
```
VadimPusiak788: Combine the separate open() context managers for source and
destination into one compound with statement to reduce nesting
and clarify that both file handles share the same scope.

VadimPusiak788 (changes requested):

app/main.py
sofiia-pohuliailo388: Changed. Please, check if everything is correct

VadimPusiak788 (approved):

PR #2278, opened 2026-08-16
solved

Aragorn-untold:
Very good


app/main.py
```diff
@@ -1 +1,11 @@
-# write your code here
+from pathlib import Path
+
+
+def copy_file(command: str) -> None:
+    command_list = command.split()
```
DonDmytron, line 5: use unpacking here for better naming

DonDmytron (changes requested):

DonDmytron (approved):
good job )

PR #2279, opened 2026-08-16
Solution

gospodenkolubomir-oss:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        return
+    if parts[0] != "cp":
+        return
```
vsmutok, line 6: Your solution works and everything is fine.

However, returning an empty `return` does not help the user understand what went wrong. I recommend returning an error or printing a message explaining what went wrong.

vsmutok (changes requested):

PR #2286, opened 2026-08-20
Solution

andriikinash127:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parts = command.split()
+    if len(parts) != 3:
+        return
+    if parts[0] != "cp":
+        return
+    if parts[1] == parts[2]:
```
Aragorn-untold: You should use unpacking with clear names instead of list indexing, so `commad, source, destination = parts` after you check for "cp" and len(parts) which you should check in one line `if len(parts) != 3 or parts[0] != "cp"`, make sure to use clear names for `if parts[1] == parts[2]` and for `with open(parts[1], "r") as file_in, open(parts[2], "w") as file_out`.

app/main.py
```diff
@@ -1 +1,13 @@
...
+        return
+    try:
+        with open(parts[1], "r") as file_in, open(parts[2], "w") as file_out:
+            file_out.write(file_in.read())
+    except FileNotFoundError:
```
Aragorn-untold: Also consider using` Path(source).exists()` from `pathlibs` for file validation. This is not a mandatory requirement, but only a suggestion.

Aragorn-untold (changes requested):

Aragorn-untold (approved):

PR #2290, opened 2026-08-21
Solution

FedirKiiko:


app/main.py
```diff
@@ -1 +1,16 @@
-# write your code here
+def copy_file(command: str) -> None:
+    parsed_command = command.split(" ")
```
DonDmytron: by default splits on spaces

app/main.py
```diff
@@ -1 +1,16 @@
...
+    if len(parsed_command) != 3 or parsed_command[0] != "cp":
+        return
+    com, source, destination = parsed_command
+    if source == destination:
```
DonDmytron: Do 1 check instead of 2

DonDmytron (changes requested):

DonDmytron (approved):
ok

PR #2296, opened 2026-08-23
Solution

NikitaFilipovWork:


app/main.py
```diff
@@ -1 +1,13 @@
-# write your code here
+def copy_file(command: str) -> None:
+    list_of_com = command.split()
```
Aragorn-untold: Check `command.startswith("cp ")` before splitting to avoid unnecessary operations

app/main.py
```diff
@@ -1 +1,13 @@
...
+    if len(list_of_com) > 0 and list_of_com[0] == "cp":
```
Aragorn-untold: Check `len(list_of_com)` to be exactly 3, everything else is a wrong command

app/main.py
```diff
@@ -1 +1,13 @@
...
+        try:
+            with (open(list_of_com[1], "r") as file_in,
+                  open(list_of_com[2], "w") as file_out):
+                if file_in != file_out:
+                    file_out.write(file_in.read())
+        except FileNotFoundError:
+            print("No such file or directory.")
+        except IndexError:
```
Aragorn-untold: You do not need to check `IndexError` if `len(list_of_com)` is exactly 3

Aragorn-untold (changes requested):

Aragorn-untold (approved):

Aragorn-untold (approved):

